PR #64, opened 2020-03-01
Adding FileWatcher and shader auto-reload

adriengivry:
The `FileWatcher` class has been added to `OvTools`. This class allows
watching a directory and getting notified of any file
added/removed/modified.

This `FileWatcher` has been introduced into the `AssetBrowser` to track
shader file changes and trigger shader recompilation automatically.

ShaderReload


adriengivry:
I'm not sure about merging this one for now. File watching could be achieved using a non-polling technique (Windows events) and should also be applied to the whole `AssetBrowser` to keep track of file renames/moves/deletions...

PR #66, opened 2020-03-06
Enable vsync in editor by default.

Kubiisek:
Vertical sync isn't enabled in the editor on startup.


adriengivry:
Actually, to be consistent, we should mirror `OvGame` behaviour in `OvEditor`, which is to enable/disable vertical synchronisation at `Context` intiailization (in `OvGame::Core::Context::ctor`).

Thus, we should add `m_context.device->SetVsync(true);`  in `OvEditor::Core::Context::ctor` right after `window->MakeCurrentContext();`

PR #70, opened 2020-03-18
WIP: Replacing all tabs per spaces

adriengivry:
Tabs are poorly interpreted by some applications (GitHub, VIM...). Their
size is dependent of the application and can cause display issues.
In order to make the whole repository more readable (Via GitHub.com), we
should adopt spaces instead of tabs.
This is a breaking change, but must be done early.


carlcc:
Hi, I personally recommend you to use [clang-format](http://clang.llvm.org/docs/ClangFormat.html),  to do the course, the benifits are:
 
1. Everybody can use his own favorite editor wothout worring about the format issue.

2. The format checking can be embeded to CI to ensure all the codes committed are properly formatted.

3. clang-format has provided serveral format configurations. It is easy to setup your own configuration based on a certain one. [The built-in styles](http://clang.llvm.org/docs/ClangFormatStyleOptions.html)

4. The replacing would simple be executing a command to invoke clang-format. No need to worry about introducing a bug caused by unexpected change of code.

NOTE: Some codestyle settings may change the including order(e.g. WebKit and Google style sort the 
consecutive include-commands with alphabetic order), which may cause compile errors or even bugs in your situation. You'd better diable this setting if you want to base your style on them.

adriengivry:
Thanks for the suggestion, I’ll look for it!

adriengivry:
This PR has too much impact on the project for now. I'm closing it for now

PR #76, opened 2020-05-09
Force using dedicated GPU

### LINKED ISSUE #40: Force Overload to use the dedicated GPU
We should force Overload to run on dedicated graphics card to prevent context creation failure due to OpenGL 4.3

image

maxbrundev:
Fixing Overload failure at context creation by forcing to use the dedicated graphics card for the Editor and the Game.

Fixes #40 

Reference: https://docs.nvidia.com/gameworks/content/technologies/desktop/optimus.htm

### DIFF
--- a/Sources/Overload/OvEditor/src/OvEditor/Main.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Main.cpp
@@ -8,16 +8,15 @@
 
 #include <OvTools/Utils/String.h>
 
+#include <OvRendering/Utils/Defines.h>
+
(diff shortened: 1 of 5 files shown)


adriengivry:
Actually, after thinking about it, I'm pretty sure it would be better to have this our rendering module (`OvRendering`). The reason for this, is that if we plan to use `OvRendering` in another project, this problem won't happened again. However, I'm not sure if it would still work, but we can try having this symbol export in `OvRendering/Context/Driver.h`, as this issue is related to the graphics driver itself.

PR #83, opened 2020-06-05
Editor crashes when deleting an already deleted file/folder

zetsumi:
Fixing file/folder suppression from `AssetBrowser` when this same file/folder has already been deleted using an external tool (Ex: Windows Explorer)


adriengivry:
Just to clarify the issue fixed by this PR:

Overload isn't auto-refreshing its `AssetBrowser` when a file/folder is added/deleted/renamed externally (With another external file system, such as explorer), thus, when an item is deleted from the disk while being display in `AssetBrowser`, trying to delete it again from the "not up-to-date" `AssetBrowser` will cause a crash.

PR #84, opened 2020-06-05
Window scroll control and console log order

adriengivry:
Adding some new methods to `PanelWindow` to control the window scrolling
(`IsScrolledToBottom`, `IsScrolledToTop`, `ScrollToTop`, `ScrollToBottom`).

There is no use case for these new methods, however it has been added in
order to test console auto-scroll to bottom when new logs arrives.

`WidgetContainer` now supports reversed draw order, meaning that all of
our classes inheriting from it (Ex: `Group`), can be setup to draw its
internal widgets in reversed order. This new feature (Reversed order)
has been used with the console to draw recents log first.

### DIFF
--- a/Sources/Overload/OvUI/src/OvUI/Internal/WidgetContainer.cpp
+++ b/Sources/Overload/OvUI/src/OvUI/Internal/WidgetContainer.cpp
@@ -72,8 +72,21 @@ void OvUI::Internal::WidgetContainer::DrawWidgets()
 {
 	CollectGarbages();
 
-	for (auto& widget : m_widgets)
-		widget.first->Draw();
+    if (m_reversedDrawOrder)
+    {
+        for (auto it = m_widgets.crbegin(); it != m_widgets.crend(); ++it)
+            it->first->Draw();
+    }
+    else
+    {
+        for (const auto& widget : m_widgets)
+            widget.first->Draw();
+    }
+}
+
+void OvUI::Internal::WidgetContainer::ReverseDrawOrder(const bool reversed)
+{
+    m_reversedDrawOrder = reversed;
 }
 
 std::vector<std::pair<OvUI::Widgets::AWidget*, OvUI::Internal::EMemoryMode>>& OvUI::Internal::WidgetContainer::GetWidgets()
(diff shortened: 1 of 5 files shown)


adriengivry:
Still waiting for feedback about this feature.
Do you guys think it would be a good idea ?

Sources/Overload/OvUI/src/OvUI/Internal/WidgetContainer.cpp
```diff
@@ -72,8 +72,21 @@ void OvUI::Internal::WidgetContainer::DrawWidgets()
 {
 	CollectGarbages();
 
-	for (auto& widget : m_widgets)
-		widget.first->Draw();
+    if (m_reversedDrawOrder)
+    {
+        for (auto it = m_widgets.rbegin(); it != m_widgets.rend(); ++it)
```
litelawliet, line 77: Since it's readonly, why not use `crbgin()` and `crend()` ?

Sources/Overload/OvUI/src/OvUI/Internal/WidgetContainer.cpp
```diff
@@ -72,8 +72,21 @@ void OvUI::Internal::WidgetContainer::DrawWidgets()
...
+            it->first->Draw();
+    }
+    else
+    {
+        for (auto& widget : m_widgets)
```
litelawliet, line 82: We could add a `const` in here as well : `const auto& widget : m_widgets`

Sources/Overload/OvUI/src/OvUI/Internal/WidgetContainer.cpp
```diff
@@ -72,8 +72,21 @@ void OvUI::Internal::WidgetContainer::DrawWidgets()
...
+            widget.first->Draw();
+    }
+}
+
+void OvUI::Internal::WidgetContainer::ReverseDrawOrder(bool reversed)
```
litelawliet, line 87: `bool reversed` could have a const as well : `const bool reversed`

litelawliet:
> Still waiting for feedback about this feature.
> Do you guys think it would be a good idea ?

Otherwise I think it is a good idea. I'm up with this feature, it adds more control which is nice to have. We may probably expand that feature to group same logs in the console if that sounds good to you ?

adriengivry:
Yeah we can open another issue to add a category system to the console

litelawliet (approved):

PR #124, opened 2020-06-22
Add CMake suppport.

Kubiisek:
Opening a new PR for CMake. All I did was copy the CMake files from the Linux+CMake branch to this one. This branch builds the generator (CMake files) on Windows and Linux.

- **What your PR is bringing to Overload (Description of the changes, use cases)**
This PR adds CMake support by creating CMakeLists.txt files across the projects. This makes it easier to port Overload to other platforms.

- **Requirements (CMake version)**
Minimum CMake version is 3.13.

- **Limitations (Supported configurations)**
Overload cannot be run immediately after build - it must be installed first and than launched.

- **Quick guide to test your changes**
To test this branch in Windows using Visual Studio 2019 you will have to
1. Download and install CMake from their [website](https://cmake.org/download/) or using the Visual Studio Installer (under Individual components)
2. Open it inside Visual Studio by 'Open a local folder' inside "Overload/Sources/Overload"
3. Wait for it to configure (see created out folder)
4. Build All.
5. Run the editor by clicking the arrow next to the run button and selecting 'OvEditor.exe (Install)' and running.
image
6. Everything should hopefully work.

- **How to keep CMake support up-to-date with Overload future updates**
The current structure of the CMake files for each project is structured like this (OvWindowing used for example):
OvWindowing/CMakeLists.txt - 'Main' project file - it links dependencies with current project and adds two subdirectories containing more CMake files.

OvWindowing/include/CMakeLists.txt - Adds header and inline files to current project. If you create a new file you add it to the list here.

OvWindowing/src/CMakeLists.txt - Same as above but with source files.

The root CMake list found inside "Sources/Overload/CMakeLists.txt" sets some global variables,  includes CMake scripts to find external dependencies (see cmake folder inside the root directory), adds all sub-projects using ```add_subdirectory(OvWindowing)``` and finally installs all the projects.

Knowing all that..
- creating a new class is as simple as listing it inside include and src CMake lists.
- to create a new project you add the subdirectory inside the root CMakeLists.txt and to install TARGETS (it should be pretty obvious from what is already done)
- to add a new external library you first create a new cmake script file to "find" the library, include it inside the root CMakeLists.txt and add it to target_link_link_libraries for the project that requires it.


adriengivry:
Nice stuff here!
Could you edit your PR description to include:
- What your PR is bringing to Overload (Description of the changes, use cases)
- Requirements (CMake version)
- Limitations (Supported configurations)
- Quick guide to test your changes
- How to keep CMake support up-to-date with Overload future updates

Kubiisek:
Thanks!
I edited the PR description.

adriengivry:
Some interrogations:
- Don't you think that maintaining a Visual Studio solution (With multiple projects) + a CMake configuration is a bit tedious ? IMO, we should get rid of every visual studio related files and only keep CMake.
- Opening `Overload/Sources/Overload` with VS using `Open a local folder` works fine. Isn't generating the solution+projects file using CMake (GUI or command line) isn't the standard approach ? (I'm a complete beginner with CMake)

Couple of issues encountered so far:
- Using `Open a local folder` (Without generating solution/projects files) builds well, however, the `Build` folder (at the repository root), is containing every project build result and is very messy. It should only contain folders such as:
    - `OvAnalytics`
    - `OvTools`
    - `OvEditor`
    - ...
- Generating a VS2019 solution using CMake-gui and building sources has some problems:
    - Linking error with `OvGame`: `error LNK2019: unresolved external symbol WinMain referenced in function "int __cdecl invoke_main(void)" (?invoke_main@@YAHXZ)`
    - No build folder created
    - No post-build triggered (DLLs, resources, `layout.ini`... aren't copied. Maybe it is due to the build failing)

In general, we should keep the exact same folder hierarchy as current VS2019 projects are generating:
- `Build` folder with sub-folders for each project
- `Bin` and `Bin-Int` for compilation generated objects

Kubiisek:
> * Don't you think that maintaining a Visual Studio solution (With multiple projects) + a CMake configuration is a bit tedious ? IMO, we should get rid of every visual studio related files and only keep CMake.

Yes that's tedious indeed. I wasn't sure if you wanted to keep the MSBuild project files. I'm going to remove them.

> * Opening `Overload/Sources/Overload` with VS using `Open a local folder` works fine. Isn't generating the solution+projects file using CMake (GUI or command line) isn't the standard approach ? (I'm a complete beginner with CMake)

That was the standard approach before Visual Studio 2017 which added built-in CMake support. There should be no difference in the resulting binary just the workflow is a little different. You can of course do it the old way if you like it better. 

> Using `Open a local folder` (Without generating solution/projects files) builds well, however, the `Build` folder (at the repository root), is containing every project build result and is very messy. It should only contain folders such as:...

I don't think I fully understand the purpose of this. Is this for including in other projects (so they can just include the include folder and link the .lib files in the lib folder)?



> * Generating a VS2019 solution using CMake-gui and building sources has some problems:
>   
>   * Linking error with `OvGame`: `error LNK2019: unresolved external symbol WinMain referenced in function "int __cdecl invoke_main(void)" (?invoke_main@@YAHXZ)`
>   * No build folder created
>   * No post-build triggered (DLLs, resources, `layout.ini`... aren't copied. Maybe it is due to the build failing)

That's the result of me trying to replicate how the OvGame currently builds - when it's in debug it builds with a console but in release it builds as a normal Windows application. CMake doesn't know what the user is going to build inside Visual Studio due to MSBuild being a multi-configuration build system as compared to Make or Ninja where you cannot choose how you are going to build your project. So that's why you can build Overload in Visual Studio (without generating the build files inside cmake-gui or the command line) but not inside the generated MSBuild projects. (Because VS by default builds CMake projects using Ninja and not MSBuild.) 
~~I'm afraid that that feature is going to have to be removed to ensure compatibility with MSBuild and Xcode generated using CMake.~~ Nevermind just found a fix 👍.

> In general, we should keep the exact same folder hierarchy as current VS2019 projects are generating:
> 
> * `Build` folder with sub-folders for each project
> * `Bin` and `Bin-Int` for compilation generated objects

As I said, I don't understand what's the `Build` folder for, please elaborate. And I'm sorry but separate `Bin` and `Bin-Int` folders won't be possible as that's how CMake works. Keep in mind that this build structure cannot be 1:1 as CMake needs to be compatible with other platforms and build their build systems.

Kubiisek:
> * No post-build triggered (DLLs, resources, `layout.ini`... aren't copied. Maybe it is due to the build failing)

Don't forget to build the INSTALL project inside VS.

adriengivry:
Thanks for your precisions, it is becoming way clearer for me!
If any output folder is generated during compilation, it should 100% be at the root of the repository instead of under `Overload/Sources`.

The `Build` folder point is to have something like:

image

and inside, let say `OvEditor/Debug` :
image

and `OvCore`:
image

This makes it very convenient to extract some libraries, to re-use them somewhere else.

For executable projects, the output in build should be:
- `ProjectName`
  - `Debug`
    - `ProjectName.exe`
    - `DLLs`
    - (Other files: Resources, config...)
  - `Release`
     - (Same as Debug)

For library projects:
- `ProjectName`
  - `bin`
    - `Debug`
      - `Some DLLs`
    - `Release`
      - `Some DLLs`
  - `include`
    - (headers here)
  - `lib`
    - `Debug`
       - `Some LIBs`
    - `Release`
      - `Some LIBs`

Another thing, it could be nice to automate the `add file to CMake target list` process. As I said, I'm a complete beginner with CMake but take a look at https://stackoverflow.com/questions/3201154/automatically-add-all-files-in-a-folder-to-a-target-using-cmake

Kubiisek:
Reuse certainly makes sense and it's actually super easy to reuse the subprojects with the current structure with CMake's `add_subdirectory` command you can include the subproject. And with `target_link_libraries` you can link against the project. So I don't see the need to copy the directory structure of the current build system. Of course it's doable but I don't really think it's that necessary.  And note that the debug inside VS would still run on the actual result of the compilation, not the copied one.
What I'm thinking is that CMake will copy for the ease of iteration and debugging all dependencies of the OvEditor to it's target build folder and the install command would be used to create an installer (using `CPack`) on Windows and eventually some equivalent on other platforms. How does that sound? As I said that directory structure is possible if you really want it.

File globing is generally [discouraged](https://stackoverflow.com/questions/32411963/why-is-cmake-file-glob-evil) in the CMake community so I thought it would be 'right' to list them all. However, I don't mind doing it that way.

adriengivry:
Good point about globing. We should avoid it then.

I think that generating clean build folders is essential to deploy faster and keep things simple. We should avoid generating an `OvEditor` build folder with crap in it, because it will make releases tedious and error-prone.

Kubiisek:
Just to be clear, I'm talking about DLL dependencies. MSBuild automatically copies all required DLLs for you to the target output directory. So I'm basically just recreating something that CMake for some reason doesn't do itself (you have to copy all DLLs manually in the CMake scripts). And for the required assets we can just set the current working directory of the debugger so the debugged Editor/Game can load them.

adriengivry:
Yes, we can try this. This is a big feature and we should iterate on it until we are satisfied.

Kubiisek:
I just did something I'm not really proud of but anyway, I made CMake copy the required DLLs and other runtime dependencies to OvEditor including the export feature (which by the way didn't work before). And it's a rather bad solution but by far the easiest - I copied the entire build folder with the .obj files to the Builder folder. It's a bad solution. Instead a much cleaner solution would be symlinks. We wouldn't need to delete the DLLs and then copy them and it would save disk space and imporve the build times. However, you can only create symlinks as an Administrator or you have to have developer mode enabled. So if somebody would be to compile Overload and not have one of those options, it would not run (that is, if they didn't copy them themselves). Or if creating a symlink wouldn't be possible we could maybe fallback to copying?

adriengivry:
I guess we should only copy dlls to OvEditor and ignore objs. Copying these files is really fast so we shouldn’t care about performance there.

Kubiisek:
Ok so CMake now calls cmd to copy the files. It's not a cross-platform solution but file extensions aren't either so it would have had to rewrite it for every platform anyway.
And one more thing, in the current build system you copy the debug or the release version to the coocking/builder directory but what if the release version wasn't build yet or worse it is outdated and therefore it can behave differently than the development build. I think this behavior definitely isn't wanted and the development version shouldn't be actually **not** optimized. I think it would be much better to use the release version for everything and maybe just set a bit in the executable to change it to a development version or a release one (or maybe just create a file that signifies the game it is a development version). And the development version should output any errors/warnaings/logs to a file, open a console window or use an in-game console.

adriengivry:
The problem is that the `OvGame` executable is optimized (Lightweight executable) when building in release mode (Profiling disabled, less UI panels, less symbols). Of course we can toss away this optimization and make a generic executable, with runtime branching to enable/disable profiling, UI panels creation, etc...

However, I think it's a pretty good approach to have this seperation at compile time.

What we can do, is having `OvGame` to get compiled twice per configuration (2 times for debug and 2 times for release mode), so building the whole solution in debug mode for instance will add the development AND non-development version of `OvGame` (Compiled twice in debug mode, once with the profiling system and additional UI panels, and one without).

Good work for CMake so far, I'll test your branch again soon and give you more feedback.

adriengivry:
I just tried the `Open Local Folder` with VS2019, building the whole solution works well, however, when clicking `Install` the created `Build` folder is wrong (`OvEditor` and `OvGame` are mixed up):

image

I think the `Build` folder, after installation, should contain the build result of each project (dlls, libraries, includes) as mentionned in a previous comment. So the `Build` folder should contain:
- `OvCore`
- `OvTools`
- `OvDebug`
- `OvEditor`
- etc..

And each projet having:

image

Except for `OvGame` and `OvEditor`, something like:

image

Each project should, in a post-build command, create its build folder under `Build/`.

Some other points I noticed (Correct me if I'm wrong):
- `set(CMAKE_CXX_STANDARD 17)` can be general to the whole solution instead of being set per-project.
- Generating the VS2019 solution using CMake and building works fine (Including the `Install` project), however, the working directory when launching `OvEditor` is set to the `.vcxproj` location, which should look like this:
image
which is incorrect (No `Data`, `Config` and `Builder` folder near). Morehover, having one `.sln` per generated project looks weird to me, I don't know if its the only way CMake can handle this.

Kubiisek:
The build folder is no longer the way to debug OvEditor. You no longer run the install target just the OvEditor.
image
All runtime dependencies are copied to it's build directory.
image
And note that the coocked version of the OvGame is always built in the current configuration and it's in the Development folder (just a temporary solution).
image

`set(CMAKE_CXX_STANDARD 17)` may be required if sb. includes the subproject in their's project. The separate solutions are also created because for each subproject `project()` is called and it creates a VS solution making it (more) standalone. Both of these can be removed. Your call.

I propose that we scratch the current install behavior and only do the `bin`/`include`/`lib `folders when install is triggered and not  every time the project is built. And potentially create a NSIS installer or zip/7z package with this folder structure. Creating a deploy script that builds Overload in Debug and Release configurations (so the Development/Shipping are copied) and installs it (creating the folder structure and potentially the installer/package) would also be possible.

Do you still insist on creating that folder structure on every build?

adriengivry:
Got it. I just don’t want the transition to CMake to be a regression in our release process. I want it to stay as easy as it was, otherwise the transition won’t be justified.
 
In VS2019, I tried running `OvEditor (install).exe` and `OvEditor.exe` and found that the installer was much more efficient. When modifying for instance `OvMaths` source code and right after launching `OvEditor.exe`, the changes wont get applies to the launched `OvEditor` (Latest `OvMaths` DLL won’t be copied), so I’m not sure to follow you when you say we don’t need the installer anymore to debug.

I’m ok to forget the build folder, as long as I have a way to send/upload an OvEditor release in 2 clicks.

I prefer removing the extra .sln generated by CMake.

Kubiisek:
I removed the uneeded .sln files. CMake install now creates the desired folder structure (except the distinction between Debug and Release builds in OvEditor and OvGame. That will probably be done with the help of the batch scripts).
And the DLLs not copying was also fixed.

Can you elaborate on why you think the install target is more efficient?
Do you want to create an installer or a zip file with the editor?

adriengivry:
I wrote a post in StackOverflow about some problems I encountered with VS solutions. One of these problems was the one I explained with `OvMaths` DLLs not being copied.
Take a look at this post: https://stackoverflow.com/questions/60539041/visual-studio-c-multiple-project-solution-setup?r=SearchResults

I was pretty happy when testing your CMake integration because it solved some points listed in this post (pretty much all technical limitations listed in the 2.6 section).

However, I found out that using `OvEditor.exe` instead of `OvEditor.exe (Install)` had the same limitations. Check the 2.6.2 section, it explains exactly the problem we have with VS, and the problem we also have with CMake without installing.

adriengivry:
I've tested your latest changes:
- The `Build` folder looks nice, except, as you said, that there is no `Debug` and `Release` folders (Even for libraries such as `OvCore`).
image
_There should be a `Debug` and `Release` folder into the `bin` folder_

- The problem detailed in my stack overflow post is present, and documented here:
CodeChangeNotConsidered
_Modifying `OvRendering` and playing `OvEditor` won't consider `OvRendering` changes._
- Starting `OvEditor.exe (Install)` doesn't work anymore (Working directory looks wrong, can't find data/config...)

Kubiisek:
I fixed the DLLs not copying problem. Except for the DLLs inside the cooked `OvGame` directory. Still not sure how to do that one. 

But I wanted to ask, if your reasoning behind creating that directory structure (currently the result of `INSTALL`) is to quickly zip a release and publish it (your release process). Than what's the reason for creating a folder for each project (except the editor)? You only publish the resulting editor and none of the other projects. If you wanted to use one of the projects inside something else, you could just do `add_subdirectory(OvMaths)` and voilà your project now builds `OvMaths` with it. I'm just a little confused..

Kubiisek:
I just re-read our conversation (there are some things that I missed and want to elaborate on) and I've got some things to say.
- I think that creating that folder structure is a little unnecessary if you just include the whole editor in the release.
- The reusability part is handled by cmake itself as I explained above. 
- My only real concern is the cooked directory. As it's impossible for cmake to make sure the debug/release build is up to date and copied to the required location.
- That could be fixed by creating a batch script. The script builds the project in both configurations and copies them. Sounds simple. Hopefully will be simple to implement and your release process would be simple as double clicking that script.
- **OR** just use a single executable for both Development and Shipping. That would have the downsides that you mentioned already but no batch scripts would be required. (but some work on the engine would obviously required)

Your call.

I don't think there are any other major problems. Just some code cleanup.

adriengivry:
Ok great, I'm gonna take a look at the latest version, making sure everything work great.
That's a lot of changes, obviously its hard to debate and find the right way to do it in a PR discussion.
I'll give you feedback about the current state of this PR soon.
Thanks for your help!

Kubiisek:
After a discussion with @adriengivry we decided to close this PR and focus our efforts on the engine rather then porting it to new other platforms.

Issue #114, opened 2020-06-14
Asset Metadata Editor should be able to initiate the edition of a specific asset metadata


adriengivry:
**Problem this feature should fix**
Currently, opening the Asset Metadata Editor is totally useless because in this window there is no way to select an asset to focus. The only way to edit metadata is actually right-clicking onto an asset in the Asset Browser and select `Edit metadata` which will force the Asset Metadata Editor to focus the selected asset.

**Expected solution**
Like the material editor, the Asset Metadata Editor should allow the user to select an asset to focus.

|Material Editor|Asset Metadata Editor|
|-|-|
|![image](https://user-images.githubusercontent.com/33324216/84600750-9f7f7f00-ae49-11ea-9e55-57fae453e454.png)|![image](https://user-images.githubusercontent.com/33324216/84600764-bc1bb700-ae49-11ea-920b-1cf30e9bc537.png)|

PR #126, opened 2020-06-23
Improving the AssetProperties window (Previously AssetMetadataEditor)

adriengivry:
`AssetMetadataEditor` has been renamed to `AssetProperties`.

`AssetProperties` and `MaterialEditor` are now more consistent.
Their use flow is now similar. You can either right click onto an asset
and choose to edit its metadata or edit the material if it is one OR
open the desired window (`AssetMetadataEditor` or `MaterialEditor`) and
then drag and drop an asset to this window.

The `AssetMetadataEditor` is now divided into two sections:
- Settings
- Info

The first one, "Settings", will show every metadata associated to this
asset and allow the user to modify their values.

The second one, "Info", only display information about the asset (File
type, file size, path...)

SizeConverter static class has been added to OvTools to allow size unit
conversions (Example: Bytes to Kilo Bytes, etc...).

Fixes https://github.com/adriengivry/Overload/issues/114

AssetProperties


Sources/Overload/OvTools/include/OvTools/Utils/SizeConverter.h
```diff
@@ -0,0 +1,58 @@
...
+#include <string>
+
+#include "OvTools/API/Export.h"
+
+namespace OvTools::Utils
+{
+    /*
+    * Handles size conversions
+    */
+    class API_OVTOOLS SizeConverter
```
litelawliet, line 20: Should be `final` since it will not be inherited

Sources/Overload/OvCore/src/OvCore/Helpers/GUIDrawer.cpp
```diff
@@ -291,6 +291,38 @@ OvUI::Widgets::Texts::Text& OvCore::Helpers::GUIDrawer::DrawSound(OvUI::Internal
 	return widget;
 }
 
+OvUI::Widgets::Texts::Text& OvCore::Helpers::GUIDrawer::DrawAsset(OvUI::Internal::WidgetContainer& p_root, const std::string& p_name, std::string& p_data, OvTools::Eventing::Event<>* p_updateNotifier)
+{
+    CreateTitle(p_root, p_name);
+
+    std::string displayedText = (p_data != "" ? p_data : std::string("Empty"));
```
litelawliet, line 298: I think displayedText can be const ?

litelawliet: Also performance wise, can't we just ask for p_data.empty() instead of comparing two strings ?

adriengivry: Same problem all over the `GUIDrawer` (This line is actually copied from other `GUIDrawer` methods). We can open an issue later if needed.

litelawliet (approved):

Issue #92, opened 2020-06-09
Adding `OvMaths::FMatrix4::Rotation()` and `OvMaths::FMatrix4::Rotate()`


adriengivry:
**Problem this feature should fix**
There are some inconsistencies between our matrix construction for translation, scale and rotation:
- [X] Create a translation matrix using `OvMaths::FMatrix4::Translation(translationVector)`
- [X] Create a scaling matrix using `OvMaths::FMatrix4::Scaling(scalingVector)`
- [x] Create a rotation matrix using `OvMaths::FMatrix4::Rotation(rotationQuaternion)`

Morehover, when we already have a transformation matrix, matrix manipulation is inconsistent:
- [X] Translate matrix using  `OvMaths::FMatrix4::Translate(matrix, translationVector)`
- [X] Scale matrix using  `OvMaths::FMatrix4::Scale(matrix, scalingVector)`
- [x] Rotate matrix using  `OvMaths::FMatrix4::Rotate(matrix, rotationQuaternion)`

**Expected solution**
We should implement:
- `OvMaths::FMatrix4::Rotation(rotationQuaternion)`
- `OvMaths::FMatrix4::Rotate(matrix, rotationQuaternion)`

These new methods should also get exposed to lua scripting

litelawliet:
I think this issue can be closed since it is merged and fully completed by kmqwerrty ?

adriengivry:
Nope, lua binding is still to be done

PR #130, opened 2020-09-16
adding OvMaths::FMatrix4::Rotation and OvMaths::FMatrix4::Rotate

kmqwerty:
Contributes to #92

### DIFF
--- a/Sources/Overload/OvMaths/include/OvMaths/FMatrix4.h
+++ b/Sources/Overload/OvMaths/include/OvMaths/FMatrix4.h
@@ -15,7 +15,7 @@
 #include "OvMaths/API/export.h"
 #include "OvMaths/FVector3.h"
 #include "OvMaths/FVector4.h"
-#include "OvMaths/FQuaternion.h"
+
 
 namespace OvMaths
 {
--- a/Sources/Overload/OvMaths/include/OvMaths/FMatrix4.h
+++ b/Sources/Overload/OvMaths/include/OvMaths/FMatrix4.h
@@ -15,9 +15,16 @@
 #include "OvMaths/API/export.h"
 #include "OvMaths/FVector3.h"
 #include "OvMaths/FVector4.h"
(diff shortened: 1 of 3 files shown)


Sources/Overload/OvMaths/include/OvMaths/FMatrix4.h
```diff
@@ -15,9 +15,16 @@
 #include "OvMaths/API/export.h"
 #include "OvMaths/FVector3.h"
 #include "OvMaths/FVector4.h"
+#include "OvMaths/FQuaternion.h"
```
adriengivry, line 18: If you forward declare `FQuaternion` because of circular reference, this include should get moved to the `.cpp` file.

Sources/Overload/OvMaths/include/OvMaths/FQuaternion.h
```diff
@@ -15,6 +15,11 @@
 
 namespace OvMaths
 {
+	/**
+	* Forward declaration due to circular reference
+	*/
+	struct FMatrix4;
```
adriengivry, line 21: This isn't necessary as `FMatrix4.` is already included.
It can be either removed, or we can move `#include "OvMaths/FMatrix4.h"` from this file to the `.cpp` file.

adriengivry:
Looks good overall! I've noted a couple of stuff that we should fix.

Morehover, in the future, it would be good to forward declare everything that we can in `.h` files, and move all our includes from `.h` to `.cpp` files.
This was a bad practice that I wasn't aware of at the time most of Overload code/architecture has been made. As a rule of thumb, never include other header files from a header file, unless you are implementing some inline stuff that requires knowing more than just the symbol.

Thanks for your contribution tho, it's greatly appreciated :)

adriengivry:
One other thing, this doesn't resolves #92 has this issue requires exposing these new methods to lua.

adriengivry (approved):
Looks great!

PR #131, opened 2020-09-22
Remove GL calls from anywhere else than OvRendering

### LINKED ISSUE #95: Remove GL calls from anywhere else than OvRendering
Some GL calls are still polluating various Overload modules. We should avoid any GL call outside of `OvRendering`.

Currently found:
- `SceneView.cpp:207 (glReadPixels)`
- `AView.cpp:51 (glViewport)`

If necessary, these calls must be wraps into `OvRendering` methods.

kmqwerty:
resolves #95 

glReadPixels and glViewport wrapps into Rendering methods.

glReadPixels function uses "Format" and "Type" so I add EPixelDataFormat and EPixelDataType.

I tried to see if other gl calls existed in other files else than OvRendering but as far as I know, i could not find it.

### DIFF
--- a/Sources/Overload/OvEditor/src/OvEditor/Panels/AView.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Panels/AView.cpp
@@ -50,9 +50,9 @@ void OvEditor::Panels::AView::Render()
 
 	EDITOR_CONTEXT(shapeDrawer)->SetViewProjection(m_camera.GetProjectionMatrix() * m_camera.GetViewMatrix());
 
-	auto renderer = EDITOR_CONTEXT(renderer).get();
+	auto& baseRenderer = *EDITOR_CONTEXT(renderer).get();
 
-	renderer->SetViewPort(0, 0, winWidth, winHeight);
+	baseRenderer.SetViewPort(0, 0, winWidth, winHeight);
 
 	_Render_Impl();
 }
--- a/Sources/Overload/OvEditor/src/OvEditor/Panels/AView.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Panels/AView.cpp
@@ -50,9 +50,7 @@ void OvEditor::Panels::AView::Render()
 
 	EDITOR_CONTEXT(shapeDrawer)->SetViewProjection(m_camera.GetProjectionMatrix() * m_camera.GetViewMatrix());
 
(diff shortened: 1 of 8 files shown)


Sources/Overload/OvEditor/src/OvEditor/Panels/AView.cpp
```diff
@@ -50,7 +50,9 @@ void OvEditor::Panels::AView::Render()
 
 	EDITOR_CONTEXT(shapeDrawer)->SetViewProjection(m_camera.GetProjectionMatrix() * m_camera.GetViewMatrix());
 
-	glViewport(0, 0, winWidth, winHeight); // TODO: Move this OpenGL call to OvRendering
+	auto& baseRenderer = *EDITOR_CONTEXT(renderer).get();
+
+	baseRenderer.SetViewPort(0, 0, winWidth, winHeight);
```
adriengivry, line 55: I would suggest going for the most compact:

```suggestion
	EDITOR_CONTEXT(renderer)->SetViewPort(0, 0, winWidth, winHeight);
```

Sources/Overload/OvEditor/src/OvEditor/Panels/SceneView.cpp
```diff
@@ -204,7 +204,8 @@ void OvEditor::Panels::SceneView::HandleActorPicking()
 
 		m_actorPickingFramebuffer.Bind();
 		uint8_t pixel[3];
-		glReadPixels(static_cast<int>(mouseX), static_cast<int>(mouseY), 1, 1, GL_RGB, GL_UNSIGNED_BYTE, pixel);
+		auto& baseRenderer = *EDITOR_CONTEXT(renderer).get();
+		baseRenderer.ReadPixels(static_cast<int>(mouseX), static_cast<int>(mouseY), 1, 1, OvRendering::Settings::EPixelDataFormat::RGB,OvRendering::Settings::EPixelDataType::UNSIGNED_BYTE,pixel);
```
adriengivry, line 208: Same here, I would suggest:

```suggestion
		EDITOR_CONTEXT(renderer)->ReadPixels(static_cast<int>(mouseX), static_cast<int>(mouseY), 1, 1, OvRendering::Settings::EPixelDataFormat::RGB, OvRendering::Settings::EPixelDataType::UNSIGNED_BYTE, pixel);
```

adriengivry (approved):

Issue #136, opened 2020-09-27
Fixing model import scale


adriengivry:
**Description**
Importing a model (It seems to be only affecting FBX models) in Overload often result into having big models that needs to get scaled down (to 0.01).

**To Reproduce**
1. Import an FBX
2. Add the FBX to the scene
3. Notice the issue

**Expected behavior**
FBX models should have a correct scale.

**Screenshots**
![FBXIncorrectScale](https://user-images.githubusercontent.com/33324216/94373735-445e5900-00d5-11eb-8dde-74c05698c9c8.gif)

**Resources**
This can solve the issue: https://github.com/assimp/assimp/issues/2166

PR #139, opened 2020-09-28
Fixing assimp unit scaling

adriengivry:
This PR aims to fix the wrong unit scaling issue with FBX models.

Currently, in projects made with Overload, we tend to counter-scale our objects with a 0.01 scale to compensate the centimeters to meters conversion.

The idea here is to add a flag to automatically convert these FBX using centimeters as the default unit to meters (In order to avoid manual counter-scaling).

Do note that this PR will break scenes made with previous version of Overload. If you want to convert your scene to correctly behave with this new scaling system, you'll have to disable `GLOBAL_SCALE` in your model asset properties.

This PR introduce several flags for models:
- GLOBAL_SCALE (On by default, disable it if you want your scenes created with older version to remain unchanged)
- EMBED_TEXTURES
- FORCE_GEN_NORMALS
- DROP_NORMALS
- GEN_BOUNDING_BOXES (No actual usage, as we do not care about assimp generated bounding boxes)

TODO:
- [X] Update Assimp to 5.0.1
- [X] Add Assimp 5.0.1 new flags to `EModelParserFlags`
- [x] Update engine models (Cube, Cone, Plane...)
- [x] Update editor models (Gizmo arrows, Camera...)

Fixes https://github.com/adriengivry/Overload/issues/136

AssimpUnityScalingFlag


Sources/Overload/OvEditor/src/OvEditor/Core/EditorRenderer.cpp
```diff
@@ -153,7 +153,7 @@ OvMaths::FMatrix4 OvEditor::Core::EditorRenderer::CalculateCameraModelMatrix(OvC
 {
 	auto translation = FMatrix4::Translation(p_actor.transform.GetWorldPosition());
 	auto rotation = FQuaternion::ToMatrix4(p_actor.transform.GetWorldRotation());
-	auto scale = FMatrix4::Scaling({ 0.4f, 0.4f, 0.4f });
+	auto scale = FMatrix4::Scaling({ 0.5f, 0.5f, 0.5f });
```
litelawliet, line 156: Instead of scaling the internal fbx of the editor with downscaling matrix of 1/2, we should export them to the correct scale beforehand. That line is confusing for a reader in order to understand "why" those values instead of 1

maxbrundev: This will certainly be removed with the integration of the new no-scaled editor models.

maxbrundev (approved):

Sources/Overload/OvEditor/src/OvEditor/Core/EditorRenderer.cpp
adriengivry: We discussed about it with Thibaut, and we think that we should pre-scale the model instead of manually scaling it in the `EditorRenderer`. I will update this ASAP

maxbrundev (approved):

litelawliet (approved):

PR #162, opened 2020-10-04
Improved actor creation menu

adriengivry:
The actor creation menu had duplicated code. It actually was implemented twice (In `MenuBar` and `Hierarchy`). To avoid that, an `OvEditor/Utils/` folder has been added with an `ActorCreationMenu` class. This class exposes a function that generates an actor creation menu.

It is now also possible to create a new actor by right-clicking onto the `Root` node in the hierarchy.

This PR also changes the default name of created actors.
Previously, creating an actor resulted in an actor named "New Actor".
Now, the name of the created actor depends on its type (Camera, Cube, Sphere, Physical Box...)

ActorCreationMenu

### DIFF
--- a/Sources/Overload/OvEditor/include/OvEditor/Core/EditorActions.h
+++ b/Sources/Overload/OvEditor/include/OvEditor/Core/EditorActions.h
@@ -166,8 +166,9 @@ namespace OvEditor::Core
 		* Create an empty actor
 		* @param p_focusOnCreation
 		* @param p_parent
+        * @param p_name
 		*/
-		OvCore::ECS::Actor&	CreateEmptyActor(bool p_focusOnCreation = true, OvCore::ECS::Actor* p_parent = nullptr);
+		OvCore::ECS::Actor&	CreateEmptyActor(bool p_focusOnCreation = true, OvCore::ECS::Actor* p_parent = nullptr, const std::string& p_name = "");
 
 		/**
 		* Create an actor with a model renderer and a material renderer. The model renderer with use the model identified
@@ -175,29 +176,9 @@ namespace OvEditor::Core
 		* @param p_path
 		* @param p_focusOnCreation
 		* @param p_parent
+        * @param p_name
 		*/
-		OvCore::ECS::Actor&	CreateActorWithModel(const std::string& p_path, bool p_focusOnCreation = true, OvCore::ECS::Actor* p_parent = nullptr);
-
-		/**
-		* Create an actor whith a physical box
-		* @param p_focusOnCreation
-		* @param p_parent
-		*/
-		OvCore::ECS::Actor&	CreatePhysicalBox(bool p_focusOnCreation = true, OvCore::ECS::Actor* p_parent = nullptr);
-
-		/**
-		* Create an actor whith a physical sphere
-		* @param p_focusOnCreation
-		* @param p_parent
-		*/
-		OvCore::ECS::Actor&	CreatePhysicalSphere(bool p_focusOnCreation = true, OvCore::ECS::Actor* p_parent = nullptr);
-
-		/**
-		* Create an actor whith a physical capsule
-		* @param p_focusOnCreation
-		* @param p_parent
-		*/
-		OvCore::ECS::Actor& CreatePhysicalCapsule(bool p_focusOnCreation = true, OvCore::ECS::Actor* p_parent = nullptr);
+		OvCore::ECS::Actor&	CreateActorWithModel(const std::string& p_path, bool p_focusOnCreation = true, OvCore::ECS::Actor* p_parent = nullptr, const std::string& p_name = "");
 
 		/**
 		* Destroy an actor from his scene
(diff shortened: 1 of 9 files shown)


Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -483,9 +483,10 @@ OvMaths::FVector3 OvEditor::Core::EditorActions::CalculateActorSpawnPoint(float
 	return sceneView.GetCameraPosition() + sceneView.GetCameraRotation() * OvMaths::FVector3::Forward * p_distanceToCamera;
 }
 
-OvCore::ECS::Actor & OvEditor::Core::EditorActions::CreateEmptyActor(bool p_focusOnCreation, OvCore::ECS::Actor* p_parent)
+OvCore::ECS::Actor & OvEditor::Core::EditorActions::CreateEmptyActor(bool p_focusOnCreation, OvCore::ECS::Actor* p_parent, const std::string& p_name)
 {
-	auto& instance = m_context.sceneManager.GetCurrentScene()->CreateActor();
+    const auto currentScene = m_context.sceneManager.GetCurrentScene();
```
litelawliet, line 488: `const auto currentScene` should be `auto* const currentScene` since `m_context.sceneManager.GetCurrentScene()` returns a pointer

litelawliet: `GetCurrentScene()` should be marked as const

litelawliet: Since we're at it, `CalculateActorSpawnPoint(10.0f)` line 495 can also be marked `const` as well as his parameter `float p_distanceToCamera`

Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -483,9 +483,10 @@ OvMaths::FVector3 OvEditor::Core::EditorActions::CalculateActorSpawnPoint(float
 	return sceneView.GetCameraPosition() + sceneView.GetCameraRotation() * OvMaths::FVector3::Forward * p_distanceToCamera;
 }
 
-OvCore::ECS::Actor & OvEditor::Core::EditorActions::CreateEmptyActor(bool p_focusOnCreation, OvCore::ECS::Actor* p_parent)
+OvCore::ECS::Actor & OvEditor::Core::EditorActions::CreateEmptyActor(bool p_focusOnCreation, OvCore::ECS::Actor* p_parent, const std::string& p_name)
```
litelawliet, line 486: `bool p_focusOnCreation` can be set const

Sources/Overload/OvEditor/include/OvEditor/Core/EditorActions.h
```diff
@@ -166,38 +166,19 @@ namespace OvEditor::Core
...
-		*/
-		OvCore::ECS::Actor&	CreatePhysicalSphere(bool p_focusOnCreation = true, OvCore::ECS::Actor* p_parent = nullptr);
-
-		/**
-		* Create an actor whith a physical capsule
-		* @param p_focusOnCreation
-		* @param p_parent
-		*/
-		OvCore::ECS::Actor& CreatePhysicalCapsule(bool p_focusOnCreation = true, OvCore::ECS::Actor* p_parent = nullptr);
+		OvCore::ECS::Actor&	CreateActorWithModel(const std::string& p_path, bool p_focusOnCreation = true, OvCore::ECS::Actor* p_parent = nullptr, const std::string& p_name = "");
```
litelawliet, line 181: I saw in `CreateActorWithModel` that both `m_context.modelManager[p_path]` and `m_context.materialManager[":Materials\\Default.ovmat"]` return a `T*` pointer based on `AResourceManager<T>::operator[](const string&)` overload.
We should then consider `auto* const model` and `auto* const material` accordingly instead of `auto model` and `auto material`

litelawliet: We can also set `bool p_focusOnCreation` constant

litelawliet (commented):
My comments mainly concerns readability since auto alone is mature enough to adapt itself to a pointer or reference. I think it is more readable to keep the `*`, also my comments doesn't change the bahaviour, `const auto` in this context is an `auto* const` but it just hides the true type (pointers here)

Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
adriengivry: When using `auto`, pointer types are automatically deduced, so `auto*` isn't required. Making `GetCurrentScene` is out of the scope of this PR.

Same thing about `CalculateActorSpawnPoint`.

Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
adriengivry: Out of the scope of this PR.
Do note that I'm not a fan of "consting" everything outside of pointer and references. `p_focusOnCreation` here is already copied, and is local to the function.
Anyway, out of the scope

adriengivry:
@litelawliet I disagree about specifying the asterisk with `auto*`. This isn't following the conventions Overload has adopted.
If you want to suggest new coding conventions, you can create a PR modifying https://github.com/adriengivry/Overload/blob/develop/CONTRIBUTING.md

PR #170, opened 2020-10-11
Refactor service locator

demurzasty:
I refactored service locator for modern C++17.  

* Real O(1) access time
* Type-safe
* No RTTI required
* No key-value storage for services needed
* No any_cast (no additional type checking)

To not changing whole solution I decided to hide implementation in Details namespace. Anyway, if implementation is too experimental can be switched with old implementation using macro definitions without changing code somewhere else.


adriengivry (commented):
That's a nice update! However, I would add an assertion if the service hasn't been found.
Great improvement

maxbrundev (approved):

Sources/Overload/OvCore/include/OvCore/Global/ServiceLocator.h
```diff
@@ -6,15 +6,48 @@
...
+			{
+				_SERVICE = &p_service;
+			}
+
+			/**
+			 * Returns a provided service
+			 */
+			static T& Get()
+			{
+				return *_SERVICE;
```
adriengivry, line 43: I would add an assert here (Using `OVASSERT`) that make sure `_SERVICE` isn't `nullptr`.

Sources/Overload/OvCore/include/OvCore/Global/ServiceLocator.h
```diff
@@ -6,15 +6,48 @@
...
+		class Locator
+		{
+		public:
+			/**
+			 * Set new instance of service
+			 * @param p_service
+			 */
+			static void Set(T& p_service)
+			{
+				_SERVICE = &p_service;
```
adriengivry, line 35: I would add an assert here (Using `OVASSERT`) that make sure `_SERVICE` is `nullptr` (We do not want to provide a service twice).

demurzasty:
I added a .inl files for service locator. Also, I changed Assertion implementation because it doesn't work as expected. 
Assertion should be executed only on debug. Message for assertion was not used.

adriengivry (approved):
Looks good!
Overload wasn't actually using assertion, so it explains why assertion were poorly handled.
Approved :)

Issue #133, opened 2020-09-23
Improving editor camera sensitivity


adriengivry:
**Problem this feature should fix**
After using a lot of other popular game engines (Unity, Unreal Engine), going back to Overload is always weird because the camera movements (Right click and middle click camera movements) are kind of off (Uncommon sensitivity).

**Expected solution**
We should try inspiring our camera rotation and translation speeds (Right click and middle click) from other game engines, such as Unity.

PR #173, opened 2020-10-15
Increased editor camera sensitivity 

sgavil:
Increased editor camera sensitivity values in order to get a smoother feeling more similar to other common engines.
Pull request related with issue #133


adriengivry (approved):
All good 👍

Issue #185, opened 2020-11-28
Compiling in release mode fails


adriengivry:
**Description**
Compiling Overload from sources in Release mode fails:
![image](https://user-images.githubusercontent.com/33324216/100492260-7cdfda80-30f8-11eb-99ad-c5e7ad4e97a0.png)

**To Reproduce**
1. Clone Overload
2. Build in Release mode
3. See error

**Expected behavior**
Release build should work

This issue has been introduced by https://github.com/adriengivry/Overload/pull/152

PR #188, opened 2021-04-11
Introducing premake in Overload

adriengivry:
This PR aims to introduce premake as our project configuration solution. It also changes engine modules from dynamically linked libraries to statically linked ones.

Fixes https://github.com/adriengivry/Overload/issues/185


litelawliet (commented):
All good and fixing https://github.com/adriengivry/Overload/issues/185 as intended.

PR #194, opened 2021-05-21
Adding BuildAll.bat script

adriengivry:
Adding a script (BuildAll.bat) triggering msbuild to build the whole
solution in debug and release mode.

### DIFF
--- a/BuildAll.bat
+++ b/BuildAll.bat
@@ -1,0 +1,19 @@
+@ECHO OFF
+
+set SOLUTION_FILE="Sources\Overload\Overload.sln"
+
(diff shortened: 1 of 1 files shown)


maxbrundev:
I have not been able to build the solution, the first condition `IF exist %SOLUTION_FILE%` fail, even if I set the whole path I trigger an MSBUILD Error: `MSBUILD: error MSB1009: Project file does not exist.` on my side.

The solution has been generated before the process of course

adriengivry:
@maxbrundev Yeah it looks like it doesn't work for everyone. We should try to find another solution that is more universal. I'm gonna close this merge request and create an issue for it.

Issue #197, opened 2021-11-08
When a new project is created, the program directly reports an error


wangdemon:
When I create a new project in Overload, it reports an error.
![error](https://user-images.githubusercontent.com/20681339/140773402-7002caa8-672a-4cf8-81bf-29b5b5c087d4.png)
My laptop hardware is poor. CPU is i5-3230M, Graphics card is NVIDIA 630. But the Graphics card 630 supports OpenGL 4.5.
![info](https://user-images.githubusercontent.com/20681339/140774259-19129a39-1d81-4956-bd6f-65a5971217c9.png)

adriengivry:
Interesting one. Is your discrete graphics card set as the default graphics processor to use in your Nvidia Configuration Panel?

wangdemon:
Yes, I have set the NVIDIA graphics card as the default graphics display card.I also set the default NVIDIA graphics card when the Overload program starts.
![无标题](https://user-images.githubusercontent.com/20681339/141437149-2ff0d03c-b3b7-4cc5-9a79-38560f2bb05f.png)

wangdemon:
I found that the problems described above do not occur when I use a desktop computer. The graphics card of my desktop computer is amd 560. So I think Overload needs a better performance graphics card to run.

adriengivry:
From what I can see online, this GPU (GeForce GT 635M) only supports OpenGL 4.2, while Overload requires OpenGL 4.3 or higher. I think we should add a popup window warning the user when specifications aren't met.

ShigotoShoujin:
Greetings, a fix is proposed in the follow pull request https://github.com/adriengivry/Overload/pull/198#issue-779979293
Please review and suggest any improvements.

wangdemon:
OK, I understand. Thank you.

PR #198, opened 2021-11-13
Intercept glfw init error and display OpenGL version requirements

ShigotoShoujin:
Issue #197 

### Current situation
On some machine without modern enough 3d acceleration, `OvWindowing::Window::CreateGlfwWindow` will throw and the user will get a generic `std::runtime_error` that does not contain the glfw error detail.

```
m_glfwWindow = glfwCreateWindow(static_cast<int>(m_size.first), static_cast<int>(m_size.second), m_title.c_str(), selectedMonitor, nullptr);

if (!m_glfwWindow)
{
	throw std::runtime_error("Failed to create GLFW window");
}
```

### How to reproduce
Start the editor in a virtual machine without 3d acceleration support. Reproduced on Windows 10 in VMWare workstation.

### Suggested solution
The goal is to tell the user about what happened, with minimal impact and risk.

In OvEditor/Main.cpp, `OvEditor::Core::Application app(projectPath, projectName);` is now inside a `TryRun` function.
`TryRun` will
- Add a listener to event `OvWindowing::Context::Device::ErrorEvent`
- Run the `OvEditor::Core::Application` constructor
- Remove the event listener
- Run the application `app->Run();`

### Tests
The fix was tested in Debug and Release mode, both in the Virtual Machine and in a real Windows 10 setup where the error does not occur.

### Screenshot
image

### DIFF
--- a/Sources/Overload/OvEditor/src/OvEditor/Main.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Main.cpp
@@ -13,6 +13,8 @@
 #include "OvEditor/Core/ProjectHub.h"
 #include "OvEditor/Core/Application.h"
 
+#include <OvWindowing/Dialogs/MessageBox.h>
+
 #undef APIENTRY
 #include "Windows.h"
 
@@ -32,6 +34,7 @@ void UpdateWorkingDirectory(const std::string& p_executablePath)
 }
 
 int main(int argc, char** argv);
+void TryRun(std::string projectPath, std::string projectName);
 
 #ifndef _DEBUG
 INT WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, PSTR lpCmdLine, INT nCmdShow)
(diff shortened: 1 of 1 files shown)


Sources/Overload/OvEditor/src/OvEditor/Main.cpp
```diff
@@ -74,10 +77,30 @@ int main(int argc, char** argv)
 	}
 
 	if (ready)
+		TryRun(projectPath, projectName);
+
+	return EXIT_SUCCESS;
+}
+
+void TryRun(std::string projectPath, std::string projectName)
```
litelawliet, line 85: I would suggest the use of `const string& projectPath, const string& projectName` as parameters. Other than that, looks goot to me 👍

ShigotoShoujin: Good eye, the update is pushed.
I've also made TryRun `static` so it cannot escape from main.cpp.

litelawliet (approved):

litelawliet (approved):

PR #203, opened 2021-11-18
Steps towards cross-platform compatibility

niansa:


adriengivry:
I'm not sure I understand why we would include some `Makefile` in the repository? Aren't these supposed to get generated by Premake?

Also, what is the purpose of this PR? By the title, it seems like this is a step towards cross-platform compatibility, but this should't get merged to develop if the purpose of these changes isn't fully met (full cross-platform compatibility).

Sources/Overload/OvAnalytics/src/OvAnalytics/Profiling/Profiler.cpp
```diff
@@ -74,8 +74,19 @@ void OvAnalytics::Profiling::Profiler::Save(OvAnalytics::Profiling::ProfilerSpy&
 	__SAVE_MUTEX.lock();
 
 	/* Check if this thread is already registered */
-	if (std::find(__WORKING_THREADS.begin(), __WORKING_THREADS.end(), std::this_thread::get_id()) == __WORKING_THREADS.end())
-		__WORKING_THREADS.push_back(std::this_thread::get_id());
+	{
```
adriengivry, line 77: What is the purpose of this change?

adriengivry, line 77: What's the purpose of this change?

Sources/Overload/OvCore/premake5.lua
```diff
@@ -11,12 +11,10 @@ project "OvCore"
 	objdir (objoutdir .. "%{cfg.buildcfg}/%{prj.name}")
 	characterset ("MBCS")
 	
-	buildoptions { "/bigobj" }
```
adriengivry, line 14: This flag is required with MSVC when compiling certain file requiring a lot of memory

adriengivry (changes requested):
I'm not sure I understand why we would include some `Makefile` in the repository? Aren't these supposed to get generated by Premake?

Also, what is the purpose of this PR? By the title, it seems like this is a step towards cross-platform compatibility, but this should't get merged to develop if the purpose of these changes isn't fully met (full cross-platform compatibility).

I would split this one in multiple PRs:
- one should focus on typos and specifying C++ standards in our premake files
- one should expand our premake files to make sure they are portable and working well with Windows & Linux

Sources/Overload/OvPhysics/premake5.lua
```diff
@@ -3,7 +3,7 @@ project "OvPhysics"
 	language "C++"
 	cppdialect "C++17"
 	files { "**.h", "**.inl", "**.cpp" }
-	includedirs { "include", dependdir .. "bullet3/include", "%{wks.location}/OvDebug/include", "%{wks.location}/OvMaths/include", "%{wks.location}/OvTools/include" }
+	includedirs { "include", dependdir .. "bullet3/include", "bullet3/include/bullet", "%{wks.location}/OvDebug/include", "%{wks.location}/OvMaths/include", "%{wks.location}/OvTools/include" }
```
litelawliet, line 6: We should keep the includedir `bullet3/include` as it is to avoid confusion by pointing out directly the sub-directoy `bullet`.

Sources/Overload/OvAnalytics/src/OvAnalytics/Profiling/Profiler.cpp
niansa: std::thread::id is not hashable by C++ standard. That means that std::find does not work except if the implementation doesn't follow the standard (like MSVC).

Sources/Overload/OvCore/premake5.lua
niansa: Possible solution woulf to be to check which compiler is used and then add this flag only if required? I do not have the premake knowledge tho

Sources/Overload/OvPhysics/premake5.lua
niansa: That's a mistake by myself. I did not mean to leave that there!

Sources/Overload/OvAnalytics/src/OvAnalytics/Profiling/Profiler.cpp
niansa: > I'm not sure I understand why we would include some `Makefile` in the repository? Aren't these supposed to get generated by Premake?
> 
> Also, what is the purpose of this PR? By the title, it seems like this is a step towards cross-platform compatibility, but this should't get merged to develop if the purpose of these changes isn't fully met (full cross-platform compatibility).
> 
> I would split this one in multiple PRs:
> 
>     * one should focus on typos and specifying C++ standards in our premake files
> 
>     * one should expand our premake files to make sure they are portable and working well with Windows & Linux

I understand. This however is something I can not achieve due to lack of knowledge. However I can ease the work for someone who actually can.

The Makefiles left in there were an accident. I should add Makefile to the .gitignore

I will consider spliting this pr into multiple prs.

niansa:
Closing, because I don't see myself working on this any further.

Issue #110, opened 2020-06-14
Improve build system logs


adriengivry:
Some logs when trying to build a game with Overload are unclear or missing:
- [ ] Scene not found should be warned at building time instead of when the game launches
- [ ] Development mode error should be more explicit for Overload sources users
- [ ] Executable rename log entry is showing `.exe.exe` instead of `.exe`

PR #208, opened 2021-11-23
Improve Build System Logs

sierra-zero:
Improve the build system logs:
-  Warns at building time if 'Start Scene' path is not valid, instead of when the game launches.
- Makes Development mode error more explicit for Overload sources users
- Executable rename log no longer outputs ".exe.exe" instead of ".exe"

Addressing:
[ Improve build system logs #110 ](https://github.com/adriengivry/Overload/issues/110)

### DIFF
--- a/Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
@@ -300,7 +300,8 @@ void OvEditor::Core::EditorActions::BuildAtLocation(const std::string & p_config
 				}
 				else
 				{
-					OVLOG_ERROR("Builder folder for \"" + p_configuration + "\" not found. Verify you have compiled Engine source code in 'Release' configuration.");
+					const std::string buildConfiguration = p_configuration == "Development" ? "Debug" : "Release";
+					OVLOG_ERROR("Builder folder for \"" + p_configuration + "\" not found. Verify you have compiled Engine source code in '" + buildConfiguration + "' configuration.");
 					failed = true;
 				}
 			}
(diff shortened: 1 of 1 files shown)


Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -292,7 +300,7 @@ void OvEditor::Core::EditorActions::BuildAtLocation(const std::string & p_config
 				}
 				else
 				{
-					OVLOG_ERROR("Builder folder for \"" + p_configuration + "\" not found. (Missing \"" + builderFolder + "\")");
+					OVLOG_ERROR("Builder folder for \"" + p_configuration + "\" not found. Verify you have compiled Engine source code in 'Release' configuration.");
```
adriengivry, line 303: What if the editor has been compiled in Release mode only, and the user tries to generate a Development mode build? It looks to me like the user will get prompted an error message concerning the Release configuration not being compiled while in this case it should specify that the Development configuration hasn't been compiled

adriengivry (changes requested):

Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
sierra-zero: Requested changes made. Let me know if any further adjustments are required.

adriengivry: LGTM!

adriengivry (approved):

litelawliet (approved):

maxbrundev (approved):

Issue #209, opened 2021-11-24
Update key engine dependencies


sierra-zero:
**Problem this feature should fix**

A number of engine dependencies are using outdated versions.
Some dependencies have received stable updates that the project is not yet benefiting from.
A number of open Issues require or would benefit from these updates, such as [Gamepad Support](https://github.com/adriengivry/Overload/issues/16) and [No Diagonal Window Cursor](https://github.com/adriengivry/Overload/issues/121).
Also they will bring associated bug fixes (although of course, potential new issues).

Updates are proposed for the following dependencies:

- ~GLEW:  2.1 -> 2.2 (Jan 2021, primarily bugfixes).~
- [ ] GLFW: 3.2 -> 3.3.5 (October 2021, improvements and bugfixes).
- ~IMGUI (docking): 1.77 WIP -> 1.85 or 1.86 WIP (Oct 2021, improvements and bugfixes).~
-  ~EDIT: ASSIMP: 5.0.1 -> 5.1.1 (Nov 2021, bugfixes).~

**Expected solution**

- ~Initially, an update to GLEW to proceed first. as the version jump consists mostly of minor bugfixes and should have minimal impact.~ 
- GLFW and IMGUI to be updated concurrently due to codependency of imguiGLFW, to minimize breaking code.
- ~EDIT: ASSIMP a small increase in versions, low chance of breaking changes.~

adriengivry:
I'm a bit resistant to updating in general. If it works, don't change it. If we can benefit from a critical bug fix, a feature or a necessary performance boost, then updating a third-party can be considered. We should be careful with updates to avoid introducing instabilities here and there.

I would also like to move to a git sub-module approach in the future.

For now, we should evaluate the impact of all these updates, and see which ones are necessary (especially for the issues you mentioned).

sierra-zero:
I can appreciate your hesitancy.
Git sub-modules are great 👍 

GLFW would be required ... and updating it compiles and runs without apparent issues.

For reference:
I had thought IMGUI would be a necessary bump, but may get away with it. Requires a small number of changes.
GLEW had no obvious issues compiling or in test scenes afaik either. 
Didn't try ASSIMP ...  but would be fairly easy to test its compatibility as its just static meshes and no animation yet.

adriengivry:
What would be the benefit of each of these updates?

sierra-zero:
To be clear, i'm no longer proposing to add them to Overload.

adriengivry:
[...]

PR #210, opened 2021-11-24
Update GLEW from 2.1 to 2.2

sierra-zero:
Update GLEW dependency from version 2.1 to 2.2

Include GLEW licence file.


sierra-zero:
Addressing GLEW upgrade in Issue:
 [ Update key engine dependencies #209 ](https://github.com/adriengivry/Overload/issues/209)

No visible issues noted. Limited testing of particles, as only one or two seem functional in the demos by default, but that is the same on 'develop' as in this PR.

Issue #103, opened 2020-06-12
Component deserialization garbage value


adriengivry:
**Description**
This issue is mainly a follow-up from a note attached to https://github.com/adriengivry/Overload/issues/69.

Adding a new serialized member to an Overload component is currently a problem. When deserializing, we are not considering the case when the XML file doesn't contain a certain entry, and thus, some garbage values are sometimes loaded.

**Practical case:**
- `CFoo` is a component with `int age = 5` (`5` being its default value)
- A scene with component `CFoo` is saved (The scene XML file contains `<age>5</age>`)
- A new member is added to `CFoo` : `float toto = 2.0f` (Again, `2.0f` being the default value)
- When deserializing the previous scene, the `OnDeserialize` of `CFoo` should call:
`SetToto(DeserializeFloat(p_doc, p_node, "toto"))`
- However, `DeserializeBoolean` will return a garbage `float` if there is no `<toto>...</toto>` entry in the scene XML).
- As the scene as been saved before adding the new `toto` member, `CFoo` will be instantiated with a garbage `toto` value instead of the default `2.0f` value.

**Expected solution**
We should find another approach to avoid code like:
```
m_camera.SetNear(OvCore::Helpers::Serializer::DeserializeFloat(p_doc, p_node, "near"));
m_camera.SetFar(OvCore::Helpers::Serializer::DeserializeFloat(p_doc, p_node, "far"));
m_camera.SetClearColor(OvCore::Helpers::Serializer::DeserializeVec3(p_doc, p_node, "clear_color"));
```
We should fall-back to member default value if the XML doesn't contain the member value.

Kubiisek:
According to [this](https://leethomason.github.io/tinyxml2/classtinyxml2_1_1_x_m_l_element.html#a926357996bef633cb736e1a558419632), all `QueryTypeText` methods return a `XMLError` which can be used to determine the success of the operation. In the case that it fails, we could instead return a default value that would be passed in.
So the call would look like this:
```
m_camera.SetNear(OvCore::Helpers::Serializer::DeserializeFloat(p_doc, p_node, "near", m_defaultNear));
m_camera.SetFar(OvCore::Helpers::Serializer::DeserializeFloat(p_doc, p_node, "far", m_defaultFar));
m_camera.SetClearColor(OvCore::Helpers::Serializer::DeserializeVec3(p_doc, p_node, "clear_color", m_defaultClearColor));
```
or just pass in the initialized values:
```
[...]

PR #214, opened 2021-12-14
Deserialization Default Values

sierra-zero:
Adding error checking to serializer::deserialize functions, and adding default value fall-back when deserialization error occurs.

Addressing this issue: [ Component deserialization garbage value #103 ](https://github.com/adriengivry/Overload/issues/103)

Just let me know and i'll apply any changes required.
Could use insight on any issues/better practices with Serializer.cpp changes . . . and as in the image, was unsure if Material.cpp deserializing uniforms should just be zeroed in the event of a serialization error?

MaterialDeserialize


adriengivry (approved):
Looks good to me, I think defaulting floats to 0.0f is correct. Approved 😃

Sources/Overload/OvCore/src/OvCore/Helpers/Serializer.cpp
```diff
@@ -190,247 +184,398 @@ void OvCore::Helpers::Serializer::SerializeSound(tinyxml2::XMLDocument& p_doc, t
...
+	}
+}
+
+void OvCore::Helpers::Serializer::DeserializeString(tinyxml2::XMLDocument & p_doc, tinyxml2::XMLNode * p_node, const std::string & p_name, std::string & p_out, std::string p_default)
 {
 	if (auto element = p_node->FirstChildElement(p_name.c_str()); element)
 	{
 		const char* result = element->GetText();
-		p_out = result ? result : "";
+		if (result == nullptr)
```
maxbrundev, line 204: In the case you want to work on that again in the future, here you do not need the statement since is there already the ternary operation that check the result, then you can just replace the `""` by the default value.

PR #215, opened 2021-12-15
Revert "Deserialization Default Values"

sierra-zero:
Reverts adriengivry/Overload#214

Unfortunately not fully cooked, issues with string deserialization will break saved scenes etc.
Apologies, requires better testing.

Revert requested.


maxbrundev:
I think your issue can come from this situation where if the result is bull you set the returned value to `""`

https://github.com/adriengivry/Overload/pull/214#discussion_r770744122

PR #216, opened 2022-03-20
VS2022 Support

sierra-zero:
Add GenerateProjects.bat for vs2022, update premake binary for vs2022 support.


adriengivry (approved):
Looks good to me! We should probably also add vs2022 in our msbuild.yml action

PR #225, opened 2022-11-22
build support xmake

car520120:


adriengivry:
@car520120 We are currently using premake to generate solutions for Overload, what would be the advantage moving from premake to xmake?

adriengivry:
Do we still want to support xmake? If not, let's close this PR until we come up with a rationale for it

Issue #231, opened 2022-11-24
More flexibility for generating the solution using premake5


litelawliet:
**Problem this feature should fix**
For now we can only generate the solution with `GenerateProjects_vs2019.bat` or `GenerateProjects_vs2022.bat`. Which isn't very flexible or scalable with future release. Also most users will still be able to build and work on this engine with Visual Studio 2017 or greater (and even lower potentially).

**Expected solution**
We should provide only one .bat file to generate the solution while allowing the user to choose the IDE he wants. If no explicit IDe is given by the user, then we provide vs2022 as default.

Examples:
`GenerateProjects.bat` results in vs2022 solution generation.
`GenerateProjects.bat vs2017` results in vs2017 solution generation.

PR #232, opened 2022-11-24
Added flexbility to generate a solution

litelawliet:
Fix https://github.com/adriengivry/Overload/issues/231


adriengivry:
LGTM, we'll just need to update the README in another PR to document that

adriengivry (approved):

Issue #87, opened 2020-06-07
Object scale shouldn't influence gizmo translation


adriengivry:
**Description**
When an object is child of a scaled object, its translation speed using gizmo is affect

**To Reproduce**
1. Create an empty `GameObject`
1. Create a cube, child of the previous `GameObject`
2. Scale the empty `GameObject` to 0.1
3. Move the cube using the translation gizmo
4. Notice the issue (Slow translation)

**Expected behavior**
The translation gizmo behaviour shouldn't consider the scale of the object

adriengivry:
The problem here is that 1 unit in local space is different from 1 unit in world space. In our current implementation of Gizmos, we are moving objects in the space with unit coefficient instead of using raycasts. 

Here is how we move our translated object:
`m_target->transform.SetLocalPosition(originPosition + GetRealDirection() * translationCoefficient);`
We should have some kind of `SetWorldPosition` to bypass the local space. We should also add `SetWorldRotation` and `SetWorldScale`. I'll open another issue for this new methods to implement.

PR #244, opened 2023-05-29
Fixing UpdateLocalMatrix computation order if transform has parent + …

maxbrundev:
Fixing issue #87 where the scale of the parent transform where influencing gizmo operations and fixing wrong matrices multiplication order in the UpdateLocalMatrix method from the FTransform class if the transform has parent.


adriengivry (approved):
This indeed fixes the parent scale impact on child translation!

adriengivry (approved):
Tested and approved :)

PR #245, opened 2023-06-13
More lua libraries

SolarcGit:
Add the following libraries to the ScriptInterpreter.cpp PLEASE

package
coroutine
string
os
table

especially table. Also this is my first pull request please If I put it in the wrong place or did anything wrong let me know I am in the discord you can ping @Stunks and I will respond

Thanks!


adriengivry:
I believe you wanted to open an issue rather than create a pull request.

The pull request you created proposes to merge develop into master, which is something we do for releases. If you want to propose some code, please branch out from develop.

PR #254, opened 2023-07-06
Use implicitely our math in ImGui

litelawliet:
Depends on https://github.com/adriengivry/Overload/pull/252 (to combine changes in ImGui from the mentionned PR with this one).

This will allow us to forget about the `Converter` class in OvUI and use `ImVec2` and `ImVec4` seemlessly with our `OvMaths::FVector2` and `OvMaths::FVector4`.

The only trade-off is the Color struct. I had to make some overloads like `operator==(const OvMaths::FVector4)` and `operator=(const OvMahts::FVector4)` to allow the affectation of a color from a vector4.
We could instead use a specialized method in Color struct like `SetColor(const OvMaths::FVector4)` for example if we think it would be more appropriate and less error-prone when using Color and vector4.

Other than that, using our custom maths inside ImGui methods are seemless and works as intended.

Also added `@brief` in Color.h documentation to remove the `*` from showing up in the doc in VS.


adriengivry:
That's a good idea. Let's put this on hold until we merge its dependency

Issue #256, opened 2023-10-04
Editor Hierarchy stack overflow due to node manipulation


maxbrundev:
**Description**
Currently it's possible via the Hierarchy to attach a Parent Node to one of his current Children, this manipulation is causing a stack overflow due to multiple transform matrix computation.

**To Reproduce**
1. Create an Actor and attach multiple other Actors
2. Drag and Drop the Parent Actor to one of his children
3. Notice the stack overflow occurring

**Expected behavior**
On other Engines this Hierarchy manipulation seems avoided, we certainly should also do that.

PR #257, opened 2023-10-04
Fixing Hierarchy stack overflow due to parent node manipulation

maxbrundev:
Fixing issue #256 by adding a return statement in the DataReceivedEvent lambda if the Source is the parent of the Target + Removing unnecessary methods call since Actor SetParent will call AttachActorToParent via the AttachEvent event

### DIFF
--- a/Sources/Overload/OvCore/include/OvCore/ECS/Actor.h
+++ b/Sources/Overload/OvCore/include/OvCore/ECS/Actor.h
@@ -101,6 +101,12 @@ namespace OvCore::ECS
 		void DetachFromParent();
 
 		/**
+		* Returns true if this actor transform is descendant of the actor
+		* @param p_actor
+		*/
+		bool IsDescendantOf(const Actor* p_actor) const;
+
+		/**
 		* Returns true if the actor has a parent
 		*/
 		bool HasParent() const;
--- a/Sources/Overload/OvEditor/src/OvEditor/Panels/Hierarchy.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Panels/Hierarchy.cpp
@@ -282,6 +282,9 @@ void OvEditor::Panels::Hierarchy::AddActorByInstance(OvCore::ECS::Actor & p_acto
 	textSelectable.AddPlugin<OvUI::Plugins::DDSource<std::pair<OvCore::ECS::Actor*, OvUI::Widgets::Layout::TreeNode*>>>("Actor", "Attach to...", std::make_pair(&p_actor, &textSelectable));
 	textSelectable.AddPlugin<OvUI::Plugins::DDTarget<std::pair<OvCore::ECS::Actor*, OvUI::Widgets::Layout::TreeNode*>>>("Actor").DataReceivedEvent += [&p_actor, &textSelectable](std::pair<OvCore::ECS::Actor*, OvUI::Widgets::Layout::TreeNode*> p_element)
 	{
+		if (p_element.second == textSelectable.GetParent())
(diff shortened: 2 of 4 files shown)


Sources/Overload/OvEditor/src/OvEditor/Panels/Hierarchy.cpp
```diff
@@ -282,10 +282,8 @@ void OvEditor::Panels::Hierarchy::AddActorByInstance(OvCore::ECS::Actor & p_acto
 	textSelectable.AddPlugin<OvUI::Plugins::DDSource<std::pair<OvCore::ECS::Actor*, OvUI::Widgets::Layout::TreeNode*>>>("Actor", "Attach to...", std::make_pair(&p_actor, &textSelectable));
 	textSelectable.AddPlugin<OvUI::Plugins::DDTarget<std::pair<OvCore::ECS::Actor*, OvUI::Widgets::Layout::TreeNode*>>>("Actor").DataReceivedEvent += [&p_actor, &textSelectable](std::pair<OvCore::ECS::Actor*, OvUI::Widgets::Layout::TreeNode*> p_element)
 	{
-		if (p_element.second->HasParent())
-			p_element.second->GetParent()->UnconsiderWidget(*p_element.second);
-
-		textSelectable.ConsiderWidget(*p_element.second);
+		if (p_actor.IsDescendantOf(p_element.first))
+			return;
```
adriengivry, line 286: nit: instead of returning if the drop destination is a descendant of the drag source, we could perform the parenting if the drop destination is not a descendant of the drag source:

```
if (!p_actor.IsDescendantOf(p_element.first))
{
    p_element.first->SetParent(p_actor);
}
```

Could be nice to warn the user about why the operation is being aborted, so maybe:

```
if (!p_actor.IsDescendantOf(p_element.first))
{
    p_element.first->SetParent(p_actor);
}
else
{
    OVLOG_WARNING("Cannot attach \"" + p_element.first->GetName() + "\" to \"" + p_actor.GetName() + "\" because it is a descendant of the latter.");
}
```

Sources/Overload/OvCore/include/OvCore/ECS/Actor.h
```diff
@@ -100,6 +100,12 @@ namespace OvCore::ECS
 		*/
 		void DetachFromParent();
 
+		/**
+		* Returns true if this actor transform is descendant of the actor
+		* @param p_actor
+		*/
+		bool IsDescendantOf(const Actor* p_actor) const;
```
adriengivry, line 107: I believe this should be exposed to Lua through the `LuaActorBinder`.
The wiki should also be updated to include it.

adriengivry (changes requested):
The change looks good, was definitely needed. We might want to expose `IsDescendantOf` to Lua scripts and add some warning message to avoid user confusion.

Sources/Overload/OvCore/include/OvCore/ECS/Actor.h
maxbrundev: Comment addressed in the latest commit, I will update the wiki when the MR will be approved.

Sources/Overload/OvEditor/src/OvEditor/Panels/Hierarchy.cpp
maxbrundev: Warning added in the latest commit.

adriengivry (approved):
Looks good to me! I'll let you update: https://github.com/adriengivry/Overload/wiki/Actor

PR #261, opened 2023-10-13
Bump GLFW3

litelawliet:
Bump GLFW3 to current latest 3eaf1255b29fdf5c2895856c7be7d7185ef2b241. This will be the starting point of the next bump of imgui (with expected bugs regarding inputs).


adriengivry:
Is there any regression we should be aware of?

Also, what's the benefit of this upgrade, are there any feature in 3.4 that we might want to use? Or is this version addressing some bugs or performance issues that might impact us?

litelawliet:
No regression to be aware of. Looking at the commit history from the current version (3.3.5) to latest master show a bunch of fixes for Win32 (aside from a ton of commits regarding Wayland...). It appears to only be hotfixes in this latest release. There are no breaking changes.

However, this bump provide more functionnality than 3.3.4 and is needed to provide the diagonal arrow (see https://github.com/adriengivry/Overload/issues/121). This is to be use alongside an imgui bump as well. Both needs to work together since imgui relies on glfw to provide functionnality to use, diagonal arrow is one of the newly exposed stuff from glfw.
I did tested both bump seperatly, and code from imgui.cpp seems to corroborate my testing and assumptions so far.

adriengivry:
> No regression to be aware of. Looking at the commit history from the current version (3.3.5) to latest master show a bunch of fixes for Win32 (aside from a ton of commits regarding Wayland...). It appears to only be hotfixes in this latest release. There are no breaking changes.
> 
> However, this bump provide more functionnality than 3.3.4 and is needed to provide the diagonal arrow (see #121). This is to be use alongside an imgui bump as well. Both needs to work together since imgui relies on glfw to provide functionnality to use, diagonal arrow is one of the newly exposed stuff from glfw. I did tested both bump seperatly, and code from imgui.cpp seems to corroborate my testing and assumptions so far.

Got it, that's great. Regarding the issues we had with https://github.com/adriengivry/Overload/pull/252:
- Broken mouse wheel zoom
- Disappearing cursor
- Bug on first right click on the scene view
Is there any evidence showing that these issues might also be introduced by this PR?

litelawliet:
No found bug regarding https://github.com/adriengivry/Overload/pull/252.
However it is expected to have those bugs back when bumping imgui later on, obviously I'll try to fix those in 2 commits (1 for the bump, 1 for the input fixes). Some refactoring might be expected to fix the bugs.

adriengivry:
> No found bug regarding #252. However it is expected to have those bugs back when bumping imgui later on, obviously I'll try to fix those in 2 commits (1 for the bump, 1 for the input fixes). Some refactoring might be expected to fix the bugs.

Sounds good!

adriengivry (approved):

Issue #264, opened 2023-11-23
Switching scene while highlighting an actor causes a crash


adriengivry:
**Description**
Sometimes when switching scene while having a highlighted (not selected) actor causes a crash.

**To Reproduce**
1. Open a scene
2. Put the asset browser next to the scene view
3. Highlight an item on the border of the scene view without selecting it
4. Move your mouse cursor out of the scene view while keeping the item highlighted
5. Change scene
6. Notice the crash

**Expected behavior**
We should make sure any highlighted actor gets cleared when switching scene (or when the actor gets deleted)

**Screenshots**
![crash-highlighted-actor](https://github.com/adriengivry/Overload/assets/33324216/b75c3ce9-cc1f-4b23-a95f-34f99f0e0dff)

**Callstack**
```
>	OvEditor.exe!OvCore::ECS::Actor::IsActive() Line 104	C++
 	OvEditor.exe!OvCore::ECS::Actor::IsActive() Line 104	C++
 	OvEditor.exe!OvEditor::Core::EditorRenderer::RenderActorOutlinePass(OvCore::ECS::Actor & p_actor, bool p_toStencil, bool p_isSelected) Line 374	C++
 	OvEditor.exe!OvEditor::Panels::SceneView::RenderScene(unsigned char p_defaultRenderState) Line 151	C++
 	OvEditor.exe!OvEditor::Panels::SceneView::_Render_Impl() Line 82	C++
 	OvEditor.exe!OvEditor::Panels::AView::Render() Line 56	C++
 	OvEditor.exe!OvEditor::Core::Editor::RenderViews(float p_deltaTime) Line 229	C++
 	OvEditor.exe!OvEditor::Core::Editor::Update(float p_deltaTime) Line 89	C++
 	OvEditor.exe!OvEditor::Core::Application::Run() Line 29	C++
 	OvEditor.exe!TryRun(const std::string & projectPath, const std::string & projectName) Line 103	C++
 	OvEditor.exe!main(int argc, char * * argv) Line 80	C++
 	[External Code]	
```

PR #265, opened 2023-11-23
Fixed a crash when changing scene while highlighting an actor

adriengivry:
# Description
Added a check to clear the currently highlighted actor if this actor gets deleted

# Related Issues
Fixes https://github.com/adriengivry/Overload/issues/264

### DIFF
--- a/Sources/Overload/OvEditor/src/OvEditor/Panels/SceneView.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Panels/SceneView.cpp
@@ -33,6 +33,14 @@ OvEditor::Panels::SceneView::SceneView
 		case OvTools::Utils::PathParser::EFileType::MODEL:	EDITOR_EXEC(CreateActorWithModel(path, true));	break;
 		}
 	};
+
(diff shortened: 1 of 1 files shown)


litelawliet:
We could remove the highlight anytime we move over another window.
Highlight is a real-time feature depending of the current cursor position, I don't think keeping it highlighted when hovering on another window gives us anything interesting.
However I can see interest when highlighting from the actor hierarchy with mouse hover.

What do you think ?

adriengivry:
@litelawliet This would only hide the issue and would re-introduce it if we ever add a feature to switch scene with a keyboard shortcut while highlighting an actor

litelawliet:
I think it's complementary, functionnaly there are (to me) no advantages to keep the highlight if we're outside the view scene doing something else; so it makes sense in normal use condition to remove the highlight in such cases.
Now yes, with the shortcut we'll have to clear the scene view like the PR is doing and I think it covers it correctly for the future by doing so in `SceneView.cpp`.

It seems we can say this PR fix the issue, but it would be good to create a new issue to discard any highlight when the cursor is outside the scene view.

litelawliet (approved):

Issue #278, opened 2023-12-03
HardwareInfo Panel widgets memory leak


maxbrundev:
**Description**
The created widgets that are by default internally memory managed are never freed

**Expected behavior**
Remove all widgets in the destructor

adriengivry:
Fixed by https://github.com/adriengivry/Overload/pull/279

PR #279, opened 2023-12-03
Fixing HardwareInfo panel memory leak

maxbrundev:
Fixing issue #278

### DIFF
--- a/Sources/Overload/OvCore/src/OvCore/ECS/Components/CCamera.cpp
+++ b/Sources/Overload/OvCore/src/OvCore/ECS/Components/CCamera.cpp
@@ -143,13 +143,13 @@ void OvCore::ECS::Components::CCamera::OnInspector(OvUI::Internal::WidgetContain
     auto currentProjectionMode = GetProjectionMode();
 
 	OvCore::Helpers::GUIDrawer::DrawScalar<float>(p_root, "Field of view", std::bind(&CCamera::GetFov, this), std::bind(&CCamera::SetFov, this, std::placeholders::_1));
-    auto& fovWidget = *p_root.GetWidgets()[p_root.GetWidgets().size() - 1].first;
-    auto& fovWidgetLabel = *p_root.GetWidgets()[p_root.GetWidgets().size() - 2].first;
+    auto& fovWidget = *p_root.GetWidgets()[p_root.GetWidgets().size() - 1];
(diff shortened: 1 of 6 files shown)


adriengivry:
I just realized that the memory management of our UI elements is not super great. I can see that we never remove widgets properly. To be honest we should never have to remove widgets manually in the destructor, but instead memory should be freed by the `WidgetContainer` on destruction (for widget with internally managed memory).

I would suggest that instead of going over all our panels and calling `RemoveAllWdigets` manually, we should update the `WidgetContainer` to properly de-allocate memory using RAII.

adriengivry (commented):
I'd suggest looking to solve `OvUI` de-allocation issues instead of addressing de-allocation for each panel manually

maxbrundev:
> I just realized that the memory management of our UI elements is not super great. I can see that we never remove widgets properly. To be honest we should never have to remove widgets manually in the destructor, but instead memory should be freed by the `WidgetContainer` on destruction (for widget with internally managed memory).
> 
> I would suggest that instead of going over all our panels and calling `RemoveAllWdigets` manually, we should update the `WidgetContainer` to properly de-allocate memory using RAII.

Indeed! I updated the WidgetContainer with shared_ptr due to the possibility to consider not owned AWidget instances

adriengivry (commented):
Your removal of the `p_manageMemory` option (which allow for external memory management of widgets) got me thinking if we ever really needed it.

There is one use case in the `FPSCounter`, which is unconventional, as the widget is allocated on the stack (because it's in the `FPSCounter` class member definition), and the reason we did that was because we wanted to make it available to the `Update` method.

If instead we used `CreateWidget` to create the `FPSCounter.m_text`, we could store the widget as an optional reference (`std::optional<std::reference_wrapper<TextColored>>`), which would completely remove our only use case for this feature.

Not sure your latest change fully support the external memory management case (actually I double checked and it doesn't, casting a stack allocated member variable to a `shared_ptr` won't have the behaviour we might expect), so we'd better fully remove the only use case we have (`FPSCounter`).

On another note, I would advise against using `shared_ptr` here, as you don't need to share the memory ownership. The ownership and responsibility should only be handled by the `WidgetContainer`, so `unique_ptr` would make more sense here.

adriengivry (approved):
LGTM!

maxbrundev:
Opening issue #281 in order to address the memory management rework of the Widgets in an another dedicated PR

Merging this PR in order to provide a quick fix for the memory leaks occurring from the `WidgetContainer ` usage

PR #289, opened 2023-12-10
Updated coding conventions in `CONTRIBUTING.md`

adriengivry:
Updating `CONTRIBUTING.md` to better reflect our current coding conventions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,9 +23,9 @@ In short, when you submit code changes, your submissions are understood to be un
 ## Report bugs using Github's issues
 We use GitHub issues to track public bugs. Report a bug by opening a new issue it's that easy!
 
-## Use a Consistent Coding Style
-* Interfaces starts by `I`
-* Abstracts starts by `A`
+## Coding Conventions
+* Interfaces starts with `I`
+* Abstracts starts with `A`
 * Class names: `UpperCamelCase`
 * Public member variables: `lowerCamelCase`
 * Private member variables: `m_lowerCamelCase`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,41 +1,50 @@
...
+We use GitHub issues to track bugs. Report a bug by opening a new issue, it's that easy!
+
+## Coding Conventions
+* Interfaces starts with `I`.
+* Abstracts starts with `A`.
+* Class names: `UpperCamelCase`.
+* Public member variables: `lowerCamelCase`.
+* Private member variables: `m_lowerCamelCase`.
+* Public static variables: `UpperCamelCase`.
+* Private static variables: `_CAPS_LOCK_WITH_UNDERSCORES`.
```
maxbrundev, line 35: I am not really convinced by the naming convention of the static variables with the underscore at the beginning that are usually for reserved variables, also the double underscore naming that is for example used in the Clock class from OvTools but it's not part of the convention so should be corrected soon.

Would suggest to just remove the underscore at the beginning and keeping the Caps lock with underscores

adriengivry: I agree, I didn't modify these in this pass, but I guess it's a great time to look into it. Not sure about `UPPER_SNAKE_CASE` for private static variables though. I'd like to suggest the most commonly used `s_lowerCamelCase`. What do you think?

maxbrundev: Yes let's stick to the Hungarian notation for that!

adriengivry:
Follow up task to cleanup the codebase to align with the conventions: https://github.com/adriengivry/Overload/issues/290

maxbrundev (approved):

PR #306, opened 2024-08-03
Fix/gizmos translation

maxbrundev:
# Description
Fixing Gizmos Translations

The Gizmos Translation operation was generating an increasing offset between the target and the mouse position. It's now fixed thanks to proper computations.

**Due to the Gizmos's Shader that scale the Gizmos models depending of the camera distance it seems that the mouse is not sticking to the Gizmos arrow but that's a visual effect, there is no more undesirable offset.**

# Limitations
The current implementation does not handle properly when the object is really far from the camera, the translation is stopped due to the T multiplier value's behind 0. This statement is necessary to avoid the behavior where the target is coming back from the opposite translation direction, this can be certainly fixed easily but will require a better support math side to understand exactly where is the mistake in the current translation computation.

# Screenshots
gizmos

# Related Issues
*need to be created*


x64x2 (approved):

Sources/Overload/OvEditor/src/OvEditor/Panels/SceneView.cpp
```diff
@@ -199,11 +199,12 @@ void OvEditor::Panels::SceneView::HandleActorPicking()
 
 	if (m_gizmoOperations.IsPicking())
 	{
-		auto mousePosition = EDITOR_CONTEXT(inputManager)->GetMousePosition();
-
 		auto [winWidth, winHeight] = GetSafeSize();
 
-		m_gizmoOperations.SetCurrentMouse({ static_cast<float>(mousePosition.first), static_cast<float>(mousePosition.second) });
+		ImVec2 globalMousePos = ImGui::GetMousePos();
```
adriengivry, line 204: I don't think we want to add direct ImGui calls to `OvEditor`. Maybe we could get the mouse position from the `InputManager`?

Sources/Overload/OvEditor/src/OvEditor/Panels/SceneView.cpp
```diff
@@ -199,11 +199,12 @@ void OvEditor::Panels::SceneView::HandleActorPicking()
...
+		ImVec2 relativeMousePos = ImVec2(globalMousePos.x - m_position.x, globalMousePos.y - m_position.y);
```
adriengivry, line 205: Same here, we want to avoid using ImGui math directly

Sources/Overload/OvEditor/src/OvEditor/Core/GizmoBehaviour.cpp
```diff
@@ -108,22 +110,38 @@ OvMaths::FVector2 OvEditor::Core::GizmoBehaviour::GetScreenDirection(const OvMat
...
-	auto totalDisplacement = m_currentMouse - m_originMouse;
-	auto translationCoefficient = OvMaths::FVector2::Dot(totalDisplacement, screenDirection) * unitsPerPixel;
+	OvMaths::FVector3 direction = GetRealDirection(true);
 
-	if (IsSnappedBehaviourEnabled())
+	const OvMaths::FVector3 planePoint = m_originalTransform.GetWorldPosition();
+
+	const float denom = OvMaths::FVector3::Dot(ray.second, planeNormal);
+
+	if (std::abs(denom) == 0)
```
adriengivry, line 126: Comparing a float to an int here, maybe we could check if the `abs(denom)` is `< some_small_float`:

```suggestion
	if (std::abs(denom) <= 0.001f)
```

adriengivry: Or maybe it's fine, and we could compare it with `0.0f`

Sources/Overload/OvEditor/src/OvEditor/Core/GizmoBehaviour.cpp
```diff
@@ -108,22 +110,38 @@ OvMaths::FVector2 OvEditor::Core::GizmoBehaviour::GetScreenDirection(const OvMat
...
+		return;
+
+	const float t = OvMaths::FVector3::Dot(planePoint - ray.first, planeNormal) / denom;
+
+	if (t < 0)
```
adriengivry, line 131: Same here, comparing a float with an int

Sources/Overload/OvEditor/src/OvEditor/Core/GizmoBehaviour.cpp
```diff
@@ -108,22 +110,38 @@ OvMaths::FVector2 OvEditor::Core::GizmoBehaviour::GetScreenDirection(const OvMat
...
 
-	auto screenDirection = GetScreenDirection(p_viewMatrix, p_projectionMatrix, p_viewSize);
+	const OvMaths::FVector3 planeTangent = OvMaths::FVector3::Cross(GetRealDirection(true), m_target->transform.GetWorldPosition() - m_cameraPosition);
+	const OvMaths::FVector3 planeNormal = OvMaths::FVector3::Cross(GetRealDirection(true), planeTangent);
 
-	auto totalDisplacement = m_currentMouse - m_originMouse;
-	auto translationCoefficient = OvMaths::FVector2::Dot(totalDisplacement, screenDirection) * unitsPerPixel;
+	OvMaths::FVector3 direction = GetRealDirection(true);
 
-	if (IsSnappedBehaviourEnabled())
```
adriengivry, line 121: Does this PR preserve the snapped behaviour we previously had since this line got removed?

maxbrundev: Indeed the snap behavior was not pushed, done!

Sources/Overload/OvEditor/src/OvEditor/Core/GizmoBehaviour.cpp
```diff
@@ -108,22 +110,38 @@ OvMaths::FVector2 OvEditor::Core::GizmoBehaviour::GetScreenDirection(const OvMat
...
 
-void OvEditor::Core::GizmoBehaviour::ApplyTranslation(const OvMaths::FMatrix4& p_viewMatrix, const OvMaths::FMatrix4& p_projectionMatrix, const OvMaths::FVector2& p_viewSize) const
+void OvEditor::Core::GizmoBehaviour::ApplyTranslation(const OvMaths::FMatrix4& p_viewMatrix, const OvMaths::FMatrix4& p_projectionMatrix, const OvMaths::FVector2& p_viewSize)
 {
-	auto unitsPerPixel = 0.001f * m_distanceToActor;
-	auto originPosition = m_originalTransform.GetWorldPosition();
+	auto ray = GetMouseRay(m_currentMouse, p_viewMatrix, p_projectionMatrix, p_viewSize);
 
-	auto screenDirection = GetScreenDirection(p_viewMatrix, p_projectionMatrix, p_viewSize);
+	const OvMaths::FVector3 planeTangent = OvMaths::FVector3::Cross(GetRealDirection(true), m_target->transform.GetWorldPosition() - m_cameraPosition);
```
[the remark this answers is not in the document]
maxbrundev: Indeed! I was doing some quick tests for the Mouse ray and other Gizmo's operations but now it's no more useful.

Sources/Overload/OvEditor/src/OvEditor/Core/GizmoBehaviour.cpp
```diff
@@ -13,6 +13,15 @@ float SnapValue(float p_value, float p_step)
 	return p_value - std::fmod(p_value, p_step);
 }
 
+OvMaths::FVector3 SnapValue(OvMaths::FVector3 p_value, float p_step)
+{
+	OvMaths::FVector3 result;
+	result.x = floor(p_value.x / p_step) * p_step;
+	result.y = floor(p_value.y / p_step) * p_step;
+	result.z = floor(p_value.z / p_step) * p_step;
```
adriengivry, line 21: `std::round` seems to provide more consistent result, as otherwise using `floor` makes moving in one direction "easier" than in the other.

For instance, with a snap value of 1, when starting to drag the gizmo, if the offset on a single axis is +0.1, `floor` will keep the object in place unless the positive offset is `>= 1`, while any negative offset (let's say -0.1) would snap the object to `-1`, resulting in unbalanced translation.

```suggestion
	result.x = std::round(p_value.x / p_step) * p_step;
	result.y = std::round(p_value.y / p_step) * p_step;
	result.z = std::round(p_value.z / p_step) * p_step;
```

Sources/Overload/OvEditor/include/OvEditor/Core/GizmoBehaviour.h
```diff
@@ -74,6 +75,14 @@ namespace OvEditor::Core
 		*/
 		EDirection GetDirection() const;
 
+		/**
+		* Returns the mouse origin and direction in World Space
+		* @param p_viewMatrix
+		* @param p_projectionMatrix
+		* @param p_viewSize
```
adriengivry, line 82: nit: missing `p_mousePos` doc.

```suggestion
        * @param p_mousePos
		* @param p_viewMatrix
		* @param p_projectionMatrix
		* @param p_viewSize
```

Sources/Overload/OvEditor/src/OvEditor/Core/GizmoBehaviour.cpp
maxbrundev: Indeed I did not noticed that at first sight but totally correct!

adriengivry (approved):
LGTM 👍

PR #310, opened 2024-11-08
Fixing Transform copy in StartPicking method

### LINKED ISSUE #309: Exception occurring in Debug Config when closing the program in Play Mode
**Description**
In VS Debug Configuration, If the user is closing the program in play mode after moving an Actor that have a Children with the Translation Gizmo in the Scene View during play, an exception occur:
```
OvEditor.exe!OvMaths::FMatrix4::Multiply(const OvMaths::FMatrix4 & p_left, const OvMaths::FMatrix4 & p_right) Line 213	C++

Exception thrown: read access violation.
**p_left** was 0xFFFFFFFFFFFFFFCF.
```

**To Reproduce**
1. Create 2 Actors and set as Child one of them
3. In the Scene View, move the Parent Actor with the Translation Gizmo
4. Close the program

100% repro rate.

**Expected behavior**
The program should terminate without errors.

**Call Stack**
```
OvEditor.exe!OvMaths::FMatrix4::Multiply(const OvMaths::FMatrix4 & p_left, const OvMaths::FMatrix4 & p_right) Line 213	C++
OvEditor.exe!OvMaths::FMatrix4::operator*(const OvMaths::FMatrix4 & p_other) Line 112	C++
OvEditor.exe!OvMaths::FTransform::UpdateWorldMatrix() Line 91	C++
OvEditor.exe!OvMaths::FTransform::GenerateMatricesLocal(OvMaths::FVector3 p_position, OvMaths::FQuaternion p_rotation, OvMaths::FVector3 p_scale) Line 76	C++
OvEditor.exe!OvMaths::FTransform::NotificationHandler(OvMaths::Internal::TransformNotifier::ENotification p_notification) Line 34	C++
[External Code]	
OvEditor.exe!OvMaths::Internal::TransformNotifier::NotifyChildren(OvMaths::Internal::TransformNotifier::ENotification p_notification) Line 20	C++
OvEditor.exe!OvMaths::FTransform::~FTransform() Line 18	C++
[External Code]	
OvEditor.exe!OvEditor::Core::Editor::~Editor() Line 45	C++
OvEditor.exe!OvEditor::Core::Application::~Application() Line 19	C++
[External Code]	
OvEditor.exe!TryRun(const std::string & projectPath, const std::string & projectName) Line 104	C++
OvEditor.exe!main(int argc, char * * argv) Line 78	C++
[External Code]	
```

maxbrundev:
Fixing issue #309 by avoiding Transform copy in the StartPicking method from the GizmoBehaviour class.

_The TransformNotifier of the Parent was copied and was calling the NotificationHandler method from already destroyed instance of the Children, resulting in access read violation in the matrix class when performing matrix multiplication in the UpdateWorldMatrix method due to parent dangling pointer._

Fixes #309 _(I've added this line as it enables GitHub to automatically link this PR on the issue page)_

### DIFF
--- a/Sources/Overload/OvEditor/include/OvEditor/Core/GizmoBehaviour.h
+++ b/Sources/Overload/OvEditor/include/OvEditor/Core/GizmoBehaviour.h
@@ -140,7 +140,7 @@ namespace OvEditor::Core
 		OvCore::ECS::Actor* m_target = nullptr;
 		EGizmoOperation m_currentOperation;
 		EDirection m_direction;
-		OvMaths::FTransform m_originalTransform;
+		OvMaths::FTransform* m_originalTransform;
 		OvMaths::FVector3 m_initialOffset;
 		OvMaths::FVector2 m_originMouse;
 		OvMaths::FVector2 m_currentMouse;
--- a/Sources/Overload/OvEditor/src/OvEditor/Core/GizmoBehaviour.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Core/GizmoBehaviour.cpp
@@ -35,7 +35,7 @@ void OvEditor::Core::GizmoBehaviour::StartPicking(OvCore::ECS::Actor& p_target, 
 	m_target = &p_target;
 	m_firstMouse = true;
 	m_firstPick = true;
-	m_originalTransform = p_target.transform.GetFTransform();
+	m_originalTransform = &p_target.transform.GetFTransform();
 	m_distanceToActor = OvMaths::FVector3::Distance(p_cameraPosition, m_target->transform.GetWorldPosition());
 	m_currentOperation = p_operation;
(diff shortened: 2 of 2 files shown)


adriengivry:
Nice catch! However I believe this issue is more related to the way we implemented `FTransform` than the way we use it. Copying an `FTransform` _should_ be ok, but it isn't right now because when doing so, we also copy the reference to the parent as well as the notifiers. Copying an `FTransform` should only copy the position, rotation, and scale.

Proposed changed (to merge onto your PR): https://github.com/adriengivry/Overload/pull/312

adriengivry (changes requested):
I would suggest incorporating the [proposed changes](https://github.com/adriengivry/Overload/pull/312) to your PR

adriengivry (approved):

PR #315, opened 2024-11-15
Added shadow mapping for directional lights

### LINKED ISSUE #250: Shadow casting
I feel like Shadow casting should be a must have in a game engine and I would therefore greatly appreciate Overload having it.
Overloads visuals would benefit considerably by having Shadow casting added and having it there would make lighting up your game way easier and significantly prettier.

### LINKED ISSUE #313: Add missing types to serializer, material, and shader uniforms
Some basic types aren't supporter by our serializer, material, and shader uniforms, such as Mat3 and Mat4.

**Expected solution**
We should implement these types!

adriengivry:
# Description
Implementation of shadow mapping!

- Shadow casting AND receiving is enabled on the `Default.ovmat` material by default.
- When a new material is created, these 2 settings are OFF by default.
- Directional lights (components) have a "Shadow casting" option that is OFF by default.
- Includes 2 implementations for shadows: hard-shadows, and soft-shadows/PCF (default).
- Introduces a `TextureHandle` class, which wraps an OpenGL texture ID, and can be passed as a uniform. This `TextureHandle` class is used by the framebuffer to expose its rendered image.
- `Texture` now inherits from `TextureHandle`.
- `ShadowRenderPass` added: handles the rendering of shadow casters in the shadow map
- `ShadowRenderFeature` added: ensures that the shadow map is properly bound to each entity set to receive shadows.
- Shadow area size is exposed through the `CDirectionaLight` component. Changing this setting will change the radius of the shadow (bigger radius = less precision)
- Shadow area size is exposed through the `CDirectionaLight` component.
- Shadow follow camera is a setting exposed through the `CDirectionaLight` component allowing the directional light to use the position of the camera as a source for its light space matrix (for shadow map rendering), making shadows "follow" the camera around, instead of being static.
- Added a `singleUse` parameter to `Material::Set`, for properties that shouldn't be stored after drawing (i.e. shadow maps texture handles)

# Notes
This PR also updates `OvEditor` and `OvGame` premake files to disable `fastuptodate`, which forces both project to trigger their post-build command before any run (resulting in resources being properly updated between runs, even if no source file changed).

# Limitations
- Only one directional light can cast shadows per scene, if more are found, warnings will be logged in the console and the subsequent lights will be ignored.
- Shadow casting passes use a custom `Shadow.ovfx` shader, which doesn't vertex transformations outside of the regular `modelMatrix * vec4(geo_Pos, 1.0);`, which could result in shaders using instancing or vertex displacement to not cast shadows properly. One way to mitigate this would be to add a way to specify a shader override for certain passes (like shadows), through the material. However this change should be made in a separate PR.
- Shadow map resolution isn't exposed at the moment, as changing it could introduce shadow acne (the shadow bias being hard coded, not depending on the shadow map resolution).

# To-Do
- [x] Move shadow implementation outside of the editor (Rendering/Core)
- [x] Cleanup shader code
- [x] Make it use scene directional lights instead of the manually set one
- [x] Take ambient light into account
- [x] Add options to enable/disable shadow casting/receiving
- [x] Make sure the `ShadowRenderFeature` only runs on materials set to receive shadow
- [x] Cleanup frame buffer
- [x] Investigate better architecture for `Texture`/`TextureHandle`
- [x] Increase directional light radius
- [x] Make it work with other lighting models (PBR...)
- [x] Fix material exception on edit for the 2nd time (related to `std::any` usage)
- [x] Make shadow area size changeable
- [x] Validate frustum behaviour (this feature is broken anyway, so I guess we'll see later)
- [x] Add prefix to exclude some uniforms from being exposed to the material editor
- [x] Expose new methods & settings to lua
- [ ] Add new methods & settings to the wiki

# Related Issues
Fixes #250 
Fixes #313

# Screenshots
image
overload-shadow-on-off

# Resources
Mostly followed:
https://learnopengl.com/Advanced-Lighting/Shadows/Shadow-Mapping

### DIFF
--- a/Sources/Overload/OvCore/include/OvCore/Rendering/ShadowRenderPass.h
+++ b/Sources/Overload/OvCore/include/OvCore/Rendering/ShadowRenderPass.h
@@ -17,25 +17,25 @@
 #include <OvCore/ECS/Components/CAmbientSphereLight.h>
 #include <OvCore/Rendering/SceneRenderer.h>
 
-#include "OvEditor/Core/Context.h"
-#include "OvEditor/Core/GizmoBehaviour.h"
-
-namespace OvEditor::Rendering
+namespace OvCore::Rendering
 {
 	/**
 	* Draw the scene for actor picking
 	*/
-	class ShadowRenderFeature : public OvRendering::Features::ARenderFeature
+	class ShadowRenderPass : public OvRendering::Core::ARenderPass
 	{
 	public:
 		/**
 		* Constructor
 		* @param p_renderer
 		*/
-		ShadowRenderFeature(OvRendering::Core::CompositeRenderer& p_renderer);
+		ShadowRenderPass(OvRendering::Core::CompositeRenderer& p_renderer);
 
-	protected:
-		virtual void OnBeforeDraw(OvRendering::Data::PipelineState& p_pso, const OvRendering::Entities::Drawable& p_drawable);
-		virtual void OnAfterDraw(OvRendering::Data::PipelineState& p_pso, const OvRendering::Entities::Drawable& p_drawable);
+	private:
+		virtual void Draw(OvRendering::Data::PipelineState p_pso) override;
+		void DrawOpaques(OvRendering::Data::PipelineState p_pso, OvCore::SceneSystem::Scene& p_scene);
+
+	private:
+		OvCore::Resources::Material m_opaqueMaterial;
 	};
 }
--- a/Sources/Overload/OvCore/include/OvCore/Rendering/ShadowRenderPass.h
+++ b/Sources/Overload/OvCore/include/OvCore/Rendering/ShadowRenderPass.h
@@ -36,6 +36,6 @@ namespace OvCore::Rendering
 		void DrawOpaques(OvRendering::Data::PipelineState p_pso, OvCore::SceneSystem::Scene& p_scene);
 
 	private:
-		OvCore::Resources::Material m_opaqueMaterial;
+		OvCore::Resources::Material m_shadowMaterial;
 	};
 }
--- a/Sources/Overload/OvCore/include/OvCore/Rendering/ShadowRenderPass.h
+++ b/Sources/Overload/OvCore/include/OvCore/Rendering/ShadowRenderPass.h
@@ -20,7 +20,7 @@
 namespace OvCore::Rendering
 {
 	/**
-	* Draw the scene for actor picking
+	* Draw the scene to a depth buffer from the point of view of each light source
 	*/
 	class ShadowRenderPass : public OvRendering::Core::ARenderPass
 	{
--- a/Sources/Overload/OvCore/src/OvCore/ECS/Components/CDirectionalLight.cpp
+++ b/Sources/Overload/OvCore/src/OvCore/ECS/Components/CDirectionalLight.cpp
@@ -7,6 +7,7 @@
 #include <OvUI/Widgets/Texts/Text.h>
 #include <OvUI/Widgets/Drags/DragFloat.h>
 #include <OvUI/Widgets/Selection/ColorEdit.h>
+#include <OvUI/Widgets/Selection/ComboBox.h>
 
 #include "OvCore/ECS/Actor.h"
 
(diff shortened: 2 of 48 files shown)


adriengivry (commented):
Getting some weird logs in the console with the shadow mapping. I need to investigate!

```
[ERROR] 2024-11-21_12-34-55 OpenGL Debug Message:
Debug message (1282): GL_INVALID_OPERATION error generated. Texture name does not refer to a texture object generated by OpenGL.
Source: API
Type: Error
Severity: High
```

**Edit: Fixed! Introduced `singleUse` parameter in `Material::Set`**

Sources/Overload/OvCore/include/OvCore/Rendering/ShadowRenderPass.h
```diff
@@ -0,0 +1,41 @@
...
+#include <OvCore/ECS/Components/CModelRenderer.h>
+#include <OvCore/Resources/Material.h>
+#include <OvCore/ECS/Components/CAmbientBoxLight.h>
+#include <OvCore/ECS/Components/CAmbientSphereLight.h>
+#include <OvCore/Rendering/SceneRenderer.h>
+
+namespace OvCore::Rendering
+{
+	/**
+	* Draw the scene for actor picking
```
adriengivry, line 23: Wrong comment

Sources/Overload/OvCore/src/OvCore/Rendering/ShadowRenderPass.cpp
```diff
@@ -0,0 +1,121 @@
...
+					m_shadowMaterial.Set("_LightSpaceMatrix", lightSpaceMatrix);
+					shadowBuffer.Bind();
+					m_renderer.SetViewport(0, 0, light.shadowMapResolution, light.shadowMapResolution);
+					m_renderer.Clear(true, true, true);
+					DrawOpaques(pso, scene);
+					shadowBuffer.Unbind();
+				}
+				else
+				{
+					// Other light types not supported!
```
litelawliet, line 70: We could output an `OVLOG_WARNING("...");` to track-down if `castShadows` is set for unsupported lights. It could be seen as a "wrong setting" for those other light types

adriengivry: The option for other lights to cast shadow isn't exposed, so unless a user write custom c++ code, this shouldn't happen. I agree that a warning wouldn't hurt tho

Sources/Overload/OvCore/src/OvCore/ECS/Components/CDirectionalLight.cpp
```diff
@@ -23,17 +24,81 @@ std::string OvCore::ECS::Components::CDirectionalLight::GetName()
...
+	
+	Helpers::GUIDrawer::CreateTitle(p_root, "Shadow Map Resolution");
+
+	auto& shadowMapResolution = p_root.CreateWidget<OvUI::Widgets::Selection::ComboBox>(m_data.shadowMapResolution);
+	shadowMapResolution.choices.emplace(512, "512");
+	shadowMapResolution.choices.emplace(1024, "1024");
+	shadowMapResolution.choices.emplace(2048, "2048");
+	shadowMapResolution.choices.emplace(4096, "4096");
+	shadowMapResolution.choices.emplace(8192, "8192");
+	shadowMapResolution.choices.emplace(16384, "16384 (Experimental)");
```
litelawliet, line 100: Could be replace by a single-line to prevent multiple emplace:
```C++
	shadowMapResolution.choices = {
		{512, "512"},
		{1024, "1024"},
		{2048, "2048"},
		{4096, "4096"},
		{8192, "8192"},
		{16384, "16384 (Experimental)"}
	};
```

Sources/Overload/OvCore/src/OvCore/Rendering/ShadowRenderPass.cpp
adriengivry: Actually, if we were to add a warning here, it would log every frame, which could quickly become annoying x)

adriengivry: The other option would be to warn once the invalid setting is set, but it isn't possible since it's stored in a struct, we can't provide any validation. So... I would say we leave it as-is 😅

litelawliet (approved):

PR #316, opened 2024-11-16
Add hint to prioritize AMD discrete GPU

Synyproxy:
when AMD driver finds this variable with the appropriate value, it enables high-performance graphics mode.


adriengivry (approved):
Approved

PR #328, opened 2024-11-22
Added dynamic editor window size selection

adriengivry:
## User Problem
When working with higher resolution monitor, the default window size for Overload can be very small (1600x900). Instead of hard-coding the default resolution, it should be selected from a list of preset.

## Description
* Added `GetWorkAreaSize` method to the `Device`, allowing the editor to retrieve the "workable area" (aka safe size)
* Default window size selected to fix the highest available resolution in a pre-determined 16:9 resolutions list.
* If no available resolution is found (work area < 640, 360), an exception is thrown.

## Related Issues
N/A

### DIFF
--- a/Sources/Overload/OvEditor/src/OvEditor/Core/Context.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Core/Context.cpp
@@ -9,11 +9,57 @@
 #include <OvRendering/Entities/Light.h>
 #include <OvCore/Global/ServiceLocator.h>
 #include <OvTools/Utils/SystemCalls.h>
+#include <OvDebug/Assertion.h>
 
 #include "OvEditor/Core/Context.h"
 
 using namespace OvCore::Global;
 using namespace OvCore::ResourceManagement;
+
+constexpr std::array<std::pair<int, int>, 13> kResolutions
+{
+	std::make_pair(640, 360), // nHD
+	std::make_pair(854, 480), // FWVGA
+	std::make_pair(960, 540), // qHD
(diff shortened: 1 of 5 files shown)


Sources/Overload/OvEditor/src/OvEditor/Core/Context.cpp
```diff
@@ -8,13 +8,59 @@
...
+	std::make_pair(1600, 900), // HD+
+	std::make_pair(1920, 1080), // Full HD
+	std::make_pair(2560, 1440), // QHD
+	std::make_pair(3200, 1800), // QHD+
+	std::make_pair(3840, 2160), // 4K UHD
+	std::make_pair(5120, 2880), // 5K
+	std::make_pair(7680, 4320), // 8K UHD
+};
+
+std::array<int, 4> FindBestFitWindowSizeAndPosition(std::array<int, 4> p_workAreaSize)
```
litelawliet, line 36: Can be static

adriengivry: The only benefit of making it `static` would be to restrict this symbol to the `Context.cpp` translation unit. We haven't historically used that approach in Overload, but it could be a good way to avoid function name collisions.

The use of the `static` keyword to restrict a function to a single translation unit has been debated and even considered for deprecation; however, it is still supported as of today. I'm not sure if it's a future-proof solution.

Another way to achieve the same result (restricting the symbol to the current translation unit) is to declare the function in an anonymous namespace:
```cpp
namespace
{
    void foo() {}
}
```

litelawliet (approved):

Issue #331, opened 2024-11-25
`OvGame` runtime crash since renderer refactor


adriengivry:
**Description**
`OvGame` crashes in `CompositeRenderer::DrawFrame` due to missing output buffer (the game doesn't use framebuffers).

**To Reproduce**
1. Build the game (temp build)
2. See error

**Expected behavior**
The game shouldn't crash. The game should also make use of framebuffers now.

**Callstack**
```
OvGame.exe!OvRendering::Core::CompositeRenderer::DrawFrame() Line 43	C++
OvGame.exe!RenderCurrentScene(OvCore::Rendering::SceneRenderer & p_renderer, const OvGame::Core::Context & p_context) Line 80	C++
OvGame.exe!OvGame::Core::Game::Update(float p_deltaTime) Line 114	C++
OvGame.exe!OvGame::Core::Application::Run() Line 28	C++
OvGame.exe!main() Line 22	C++
```

PR #338, opened 2024-11-29
Fix OvGame not rendering properly

### LINKED ISSUE #337: OvGame isn't rendering properly
**Description**
Since the renderer rework (#277), it seems like `OvGame` isn't rendering the game to the backbuffer properly. The new renderer relies on an explicit framebuffer to work, which `OvGame` doesn't have as it renders to the backbuffer directly.

**To Reproduce**
1. Open Overload
2. Make a temporary build
3. See error

**Expected behavior**
`OvGame` should also use a framebuffer, and finally blit this framebuffer onto the backbuffer at the end of a frame.

**Screenshots**

adriengivry:
## Description
`OvGame` isn't properly rendering the game since post-process effects have been added. It seems like drawing directly to the backbuffer isn't supported anymore, as the renderer relies on an output framebuffer.

This PR adds a framebuffer to `OvGame`, and implements a solution to blit the framebuffer onto the backbuffer.

## Related Issues
Fixes #337
Fixes https://github.com/adriengivry/Overload/issues/331


adriengivry:
Ready for review! I ended up adding a `BlitToBackBuffer` method to the `Framebuffer` class to blit its content to the backbuffer.

litelawliet (approved):

Issue #345, opened 2025-02-07
Enable frustum culling by default everywhere


adriengivry:
**Problem this feature should fix**
Editor and in-game frustum culling is disabled by default. I believe it should be ON everywhere!

**Expected solution**
Enable frustum culling everywhere

sameep-git:
Hi, I would like to help with this issue. I am new to Overload and open source. I looked around the code base and found this file. Will the change be reflected by changing these values to true?
https://github.com/adriengivry/Overload/blob/7e74b39bb346a6e02d3ff449141309acd0f4ff87/Sources/Overload/OvEditor/include/OvEditor/Settings/EditorSettings.h#L78-L80

I was not able to find any in-game settings which would affect default behavior in my quick search. Would these changes only affect the editor or would they affect in-game too?

adriengivry:
I think so! Setting these to true should address the issue!

sameep-git:
> I think so! Setting these to true should address the issue!

Great! I will start working on it.

PR #350, opened 2025-02-08
Enable Frustum Culling by default

sameep-git:
Closes: #345

1. Change `EditorSettings.h` properties which affect Frustum Culling from `false` to `true`.

### DIFF
--- a/Sources/Overload/OvEditor/include/OvEditor/Settings/EditorSettings.h
+++ b/Sources/Overload/OvEditor/include/OvEditor/Settings/EditorSettings.h
@@ -75,9 +75,9 @@ namespace OvEditor::Settings
 
 		inline static Property<bool> ShowGeometryBounds = { false };
 		inline static Property<bool> ShowLightBounds = { false };
-		inline static Property<bool> EditorFrustumGeometryCulling = { false };
-		inline static Property<bool> EditorFrustumLightCulling = { false };
-		inline static Property<bool> DebugFrustumCulling = { false };
+		inline static Property<bool> EditorFrustumGeometryCulling = { true };
+		inline static Property<bool> EditorFrustumLightCulling = { true };
+		inline static Property<bool> DebugFrustumCulling = { true };
 		inline static Property<float> LightBillboardScale = { 0.5f };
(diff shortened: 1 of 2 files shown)


adriengivry (changes requested):
Looks good to me! Would be nice if the in-game camera also had frustum culling enabled by default. This could be done by modifying the default values here:

https://github.com/adriengivry/Overload/blob/develop/Sources/Overload/OvRendering/src/OvRendering/Entities/Camera.cpp

sameep-git:
> Looks good to me! Would be nice if the in-game camera also had frustum culling enabled by default. This could be done by modifying the default values here:
> 
> https://github.com/adriengivry/Overload/blob/develop/Sources/Overload/OvRendering/src/OvRendering/Entities/Camera.cpp

Does https://github.com/adriengivry/Overload/pull/350/commits/71cd3d5cb7b062a66851cd82d7ab42585c1aca37 fix it?

Sources/Overload/OvEditor/include/OvEditor/Settings/EditorSettings.h
```diff
@@ -75,9 +75,9 @@ namespace OvEditor::Settings
 
 		inline static Property<bool> ShowGeometryBounds = { false };
 		inline static Property<bool> ShowLightBounds = { false };
-		inline static Property<bool> EditorFrustumGeometryCulling = { false };
-		inline static Property<bool> EditorFrustumLightCulling = { false };
-		inline static Property<bool> DebugFrustumCulling = { false };
+		inline static Property<bool> EditorFrustumGeometryCulling = { true };
+		inline static Property<bool> EditorFrustumLightCulling = { true };
+		inline static Property<bool> DebugFrustumCulling = { true };
```
adriengivry, line 80: This one should actually stay set to `false`. `DebugFrustumCulling` can be toggled to apply the frustum culling of the main scene camera to the scene view, and is only useful to preview how the main camera will apply frustum culling in game

adriengivry (changes requested):
Just need to update `DebugFrustumCulling` to set it back to `false` by default, otherwise I tested it and it works as expected!

adriengivry (approved):
LGTM!

Issue #291, opened 2023-12-10
`OvMaths::Internal::TransformNotifier` shouldn't be referred to in any .h file


adriengivry:
**User problem**
`TransformNotifier` being an internal component of the `FTransform` implementation, it shouldn't be publicly exposed/visible to any other part of the engine. Currently, including `FTransform` will result into `TransformNotifier` being publicly exposed.

**Expected solution**
Move `TransformNotifier` from `FTransform.h` to `FTransform.cpp` (add a private forward declaration to the `FTransform` class declaration).

sameep-git:
Hi, I would like to tackle down this issue too. Going to open a PR, let me know if there is anything incorrect with my approach.

adriengivry:
**Edit: this issue wasn't exactly clear. I believe the idea here would be to have the `Notifier` member variable made private.**

PR #351, opened 2025-02-09
Reworked `OvMaths::Internal::TransformNotifier` within FTransform.cpp and FTransform.h

sameep-git:
Closes: #291 

1. Add a private forward declaration to `Internal::TransformNotifier` declarations, and rename the private member variable to `m_notifier` from public member variable convention `Notifier`.
2. Include "Internal/TransformNotifier.h" in FTransform.cpp instead of FTransform.h

### DIFF
--- a/Sources/Overload/OvMaths/include/OvMaths/FTransform.h
+++ b/Sources/Overload/OvMaths/include/OvMaths/FTransform.h
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include "OvMaths/Internal/TransformNotifier.h"
 #include "OvMaths/FQuaternion.h"
 #include "OvMaths/FMatrix4.h"
 #include "OvMaths/FVector3.h"
@@ -218,6 +219,9 @@ namespace OvMaths
 		*/
 		FVector3 GetLocalRight() const;
 
+	public:
+		Internal::TransformNotifier::NotificationHandlerID m_notificationHandlerID;
+	
 	private:
 		void PreDecomposeWorldMatrix();
 		void PreDecomposeLocalMatrix();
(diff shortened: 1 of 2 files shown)


Sources/Overload/OvMaths/include/OvMaths/FTransform.h
```diff
@@ -6,7 +6,6 @@
 
 #pragma once
 
-#include "OvMaths/Internal/TransformNotifier.h"
```
adriengivry, line 9: This include would still be necessary, otherwise we would need to:
-  forward declare `Internal::TransformNotifier
-  use `uint64_t` instead of `Internal::TransformNotifier::NotificationHandlerID` (not recommended, as the underlying type could change, which would require to update this too)

Even though the [issue](https://github.com/adriengivry/Overload/issues/291) mentioned that we should get rid of the include and forward declare, I guess it isn't actually possible here. Let's keep it.

adriengivry (changes requested):
Let's add back the moved include to the header file.
I still believe it's a good idea to make the `Notifier` private (`m_notifier`), so let's only keep that.

Sources/Overload/OvMaths/include/OvMaths/FTransform.h
sameep-git: Got it! I changed the files according to the requests. I built locally this time too, found a windows computer to work on this :)

adriengivry (changes requested):
Sorry I didn't make this clearer, I meant we can keep both fields (`m_notifier` and `m_notificationHandlerID`) private. We just can't forward declare them 😅

sameep-git:
> Sorry I didn't make this clearer, I meant we can keep both fields (`m_notifier` and `m_notificationHandlerID`) private. We just can't forward declare them 😅

It was totally my bad. I am still getting used to using C++ more and had to read about forward declaring to understand it. Thank you for your help. I will update the code and push the changes.

adriengivry (approved):
LGTM!

Issue #296, opened 2023-12-15
Abstract the `ScriptInterpreter` to allow for multiple implementations (Lua, Python, C#)


adriengivry:
**Problem this feature should fix**
Our `ScriptInterpreter` is currently implemented to work only with Lua.

**Expected solution**
We could provide a scripting interface that could be implemented by scripting languages, and used by `OvCore` to bind functions to any scripting language. We could find functions to bind through a reflection library (https://github.com/rttrorg/rttr) or by manually registering them from OvCore. The interface would dispatch these registration calls to the selected scripting language.

PR #355, opened 2025-02-10
Reworked the scripting system to be fully agnostic

adriengivry:
## Description
This rework will allow us to plug-in any type of scripting backend, such as Python!
* Added generic TScript and TScriptEngine classes, that can be implemented through template specialization.
* Updated the asset browser and relevant code to make sure hardcoded extensions aren't used anywhere. Valid extensions are now returned by each implementation of TScriptEngine.

## Review Guidance
Tested with Lua and a new "Dummy" scripting backend (.ovscript), which does nothing, but you can still create scripts and add them as behaviour.

## To-Do
- [x] Figure out `PathParser.cpp` reference to `lua` extension (Follow up issue: https://github.com/adriengivry/Overload/issues/358)
- [x] ~~(optional) test PyBind~~

## Related Issues
Fixes https://github.com/adriengivry/Overload/issues/296


Sources/Overload/OvCore/include/OvCore/Scripting/IScriptContext.h
```diff
@@ -0,0 +1,22 @@
...
+	/**
+	* Interface for scripting backends to implement and store script contexts
+	*/
+	class IScriptContext
+	{
+	public:
+		/**
+		* Returns true if the context is valid
+		*/
+		virtual bool IsValid() const = 0;
```
litelawliet, line 20: To be truly an interface I believe we still need to make the destructor virtual with an empty body, otherwise we're pretty much abstract.

```suggestion
	public:
	        virtual ~IScriptContext() {}
		/**
		* Returns true if the context is valid
		*/
		virtual bool IsValid() const = 0;
```

Sources/Overload/OvCore/include/OvCore/Scripting/IScriptEngine.h
```diff
@@ -0,0 +1,160 @@
...
+	* This class defines the necessary methods that a scripting engine should provide
+	* to interact with the engine's components and handle various lifecycle events.
+	*/
+	class IScriptEngine
+	{
+	public:
+		/**
+		* Returns a list of valid extensions for scripts.
+		*/
+		virtual std::vector<std::string> GetValidExtensions() = 0;
```
litelawliet, line 31: Same as above:
```suggestion
	public:
	        virtual IScriptEngine() {}
	        
		/**
		* Returns a list of valid extensions for scripts.
		*/
		virtual std::vector<std::string> GetValidExtensions() = 0;
```

Sources/Overload/OvCore/include/OvCore/Scripting/IScriptContext.h
maxbrundev: Indeed, a virtual destructor is required to avoid undefined behavior; however, I would suggest using the = default keyword for clarity and efficiency:

```suggestion
	public:
		virtual ~IScriptContext() = default;
```
This approach aligns with our modern C++ philosophy by being both concise and expressive.

Sources/Overload/OvCore/include/OvCore/ECS/Components/Behaviour.h
```diff
@@ -6,9 +6,10 @@
 
 #pragma once
 
-#include <sol.hpp>
-
-#include "OvCore/ECS/Components/CPhysicalObject.h"
+#include <OvCore/ECS/Components/CPhysicalObject.h>
+#include <OvTools/Utils/OptRef.h>
+#include <OvCore/Scripting/ScriptEngine.h>
+// namespace OvCore::Scripting { class Script; }
```
adriengivry, line 12: Oops, I need to remove that

Sources/Overload/OvCore/include/OvCore/Scripting/Common/TScriptEngine.h
```diff
@@ -0,0 +1,182 @@
+/**
+* @project: Overload
+* @author: Overload Tech.
+* @licence: MIT
+*/
+
+#pragma once
+
+#include <string>
+#include <vector>
```
adriengivry, line 10: Could be removed

litelawliet (approved):

litelawliet (changes requested):
A regression appaeared, the cursor still appears in Play Mode and doesn't allow us to move around or turn around anymore due to monitor constraints.

adriengivry:
Fixed issue where `OnAwake` was called twice and `OnStart` wasn't at all

litelawliet (approved):

PR #369, opened 2025-02-14
Fixed invalid default scene when created through the asset browser

adriengivry:
## User Problem
Scenes created through the asset browser: `Create... > Scene` are "pre-generated" and "pre-serialized". However, this method forces us to update the serialized version of the default scene when our components change.
This issue led to many errors showing up in the console when creating a scene through the asset browser, and then opening it.

## Proposed Changes
This PR addresses this issue by adding some tools to create a default scene, and serialize it, when a default scene is requested by the asset browser. This new workflow also make it easier to maintain what a "default scene" should look like, as it is more consistent and shared between parts of the engine.

## Notes
Also updated the `SceneManager::m_currentScene` variable type from `Scene*` to `std::unique_ptr<Scene*>` to follow the RAII idiom.

### DIFF
--- a/Sources/Overload/OvCore/include/OvCore/SceneSystem/Scene.h
+++ b/Sources/Overload/OvCore/include/OvCore/SceneSystem/Scene.h
@@ -46,6 +46,16 @@ namespace OvCore::SceneSystem
 		~Scene();
 
 		/**
+		* Add a default camera to the scene
+		*/
+		void AddDefaultCamera();
+
+		/**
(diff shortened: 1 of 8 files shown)


litelawliet (approved):

litelawliet:
Introduced 2 regression bugs: 

1. https://github.com/adriengivry/Overload/issues/374

2. https://github.com/adriengivry/Overload/issues/375

Issue #371, opened 2025-02-14
Whitespace in a scene name crash at launch.


litelawliet:
**Description**
Starting a game built which has a whitespace in `start_scene` setting of Game.ini will result in a `"scene not found or corrupted"`. This is due to how we trim the whitespaces in `IniFile.Load()` in OvTools.

**To Reproduce**
1. Create a scene with a space in its name, example: New Scene.ovgame
2. Set the start scene of Project Settings to this scene and apply
3. Build the game

**Expected behavior**
We should rethink the way we trim whitespaces when reading settings of `Game.ini`

**Screenshots**
![Image](https://github.com/user-attachments/assets/33c00566-b3b3-4331-8f6e-5c84bb30d2cf)

![Image](https://github.com/user-attachments/assets/acc436f0-9218-4ce4-890e-a8ac7cdf9b48)

PR #372, opened 2025-02-14
Added function to trim leading/trailing whitespaces

litelawliet:
## Description
Trim static method added in `OvTools::String` and uses an struct `TrimOptions` for the trim type/direction being left or right.

## Related Issues
Fixes https://github.com/adriengivry/Overload/issues/371.

### DIFF
--- a/Sources/Overload/OvTools/include/OvTools/Utils/String.h
+++ b/Sources/Overload/OvTools/include/OvTools/Utils/String.h
@@ -13,7 +13,7 @@
 namespace OvTools::Utils
 {
 	/*
-        * Helper class for string manipulation
+	* Helper class for string manipulation
 	*/
 	class String
 	{
@@ -45,15 +45,15 @@ namespace OvTools::Utils
 		*/
 		static void ReplaceAll(std::string& p_target, const std::string& p_from, const std::string& p_to);
 
-        /**
-        * Generate a unique string satisfying the availability predicate
-        * @param p_source
-        * @param p_isAvailable (A callback that must returning true if the input string is available)
-        */
-        static std::string GenerateUnique(const std::string& p_source, std::function<bool(std::string)> p_isAvailable);
+		/**
+		* Generate a unique string satisfying the availability predicate
+		* @param p_source
+		* @param p_isAvailable (A callback that must returning true if the input string is available)
+		*/
+		static std::string GenerateUnique(const std::string& p_source, std::function<bool(std::string)> p_isAvailable);
 
 		/**
-                * Trim whitespaces from a string with the option to trim the left and right of the string by passing a TrimOptions struct.
+		* Trim whitespaces from a string with the option to trim the left and right of the string by passing a TrimOptions struct.
 		* @param p_str String to trim
 		* @param p_trimOptions The desired trim options, default is trimming left and right
 		*/
--- a/Sources/Overload/OvTools/include/OvTools/Utils/String.h
+++ b/Sources/Overload/OvTools/include/OvTools/Utils/String.h
@@ -13,16 +13,15 @@
 namespace OvTools::Utils
 {
 	/*
-	* Handle random numbers generation
+        * Helper class for string manipulation
 	*/
 	class String
 	{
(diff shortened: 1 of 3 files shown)


Sources/Overload/OvTools/include/OvTools/Utils/String.h
```diff
@@ -45,5 +52,25 @@ namespace OvTools::Utils
...
+        * Leading trim of whitespaces from a string.
+        * param p_str String to trim
+		*/
+		static void LeadingTrim(std::string& p_str);
+
+		/**
+		* Trailing trim of whitespaces from a string.
+		* param p_str String to trim
+		*/
+		static void TrailingTrim(std::string& p_str);
```
adriengivry: These don't need to be exposed in the header file. Keep them in the `.cpp` and remove `OvTools::Utils::String::` from the function name

Sources/Overload/OvTools/src/OvTools/Utils/String.cpp
```diff
@@ -78,3 +78,27 @@ std::string OvTools::Utils::String::GenerateUnique(const std::string& p_source,
...
+   if (p_trimType & ETrimType::Trailing || p_trimType & ETrimType::TwoWays)  
+   {  
+       TrailingTrim(p_str);
+   }  
+}
+
+void OvTools::Utils::String::LeadingTrim(std::string& p_str)
+{
+    p_str.erase(p_str.begin(),
+                std::find_if(p_str.begin(), p_str.end(), [](unsigned char ch) { return !std::isspace(ch); }));
```
adriengivry, line 97:
```suggestion
    toto.erase(0, toto.find_first_not_of(" \t\n\r\f\v"));
```

Sources/Overload/OvTools/src/OvTools/Utils/String.cpp
```diff
@@ -78,3 +78,27 @@ std::string OvTools::Utils::String::GenerateUnique(const std::string& p_source,
...
+}
+
+void OvTools::Utils::String::TrailingTrim(std::string& p_str)
+{
+    p_str.erase(std::find_if(p_str.rbegin(), p_str.rend(), [](unsigned char ch) { return !std::isspace(ch); }).base(),
+                p_str.end());
```
adriengivry, line 103:
```suggestion
    toto.erase(toto.find_last_not_of(" \t\n\r\f\v") + 1);
```

adriengivry (changes requested):

Sources/Overload/OvTools/src/OvTools/Utils/String.cpp
```diff
@@ -78,3 +78,27 @@ std::string OvTools::Utils::String::GenerateUnique(const std::string& p_source,
 
     return result;
 }
+
+void OvTools::Utils::String::Trim(std::string& p_str, const ETrimType p_trimType)  
+{  
+   if (p_trimType & ETrimType::Leading || p_trimType & ETrimType::TwoWays)
```
adriengivry, line 84: It seems like your enum tries to behave like flags. Therefore, `ETrimType::TwoWays` should be set to `3`, or `ETrimType::Leading | ETrimType::Trailing`.

This check would only need to be:
```cpp
if (p_trimType & ETrimType::Leading) {}
```

Sources/Overload/OvTools/include/OvTools/Utils/String.h
```diff
@@ -18,6 +18,13 @@ namespace OvTools::Utils
 	class String
 	{
 	public:
+		enum ETrimType
+		{
+			Leading = 1,
+			Trailing = 2,
+			TwoWays = 4 // Leading and Trailing
```
adriengivry, line 25: Could be set to 3, so that `ETrimType::TwoWays == ETrimType::Leading & ETrimType::Trailing`

adriengivry (approved):
LGTM. Approved conditionally, provided that the spaces/tabs issues are fixed. We should only be using tabs

PR #391, opened 2025-02-20
Fixed OvGame window size & position

adriengivry:
## Description
Fixed OvGame window size & position:
- Windows size is divided by 2 until it fits in the work area size (if the resolution is decreased, a warning is displayed)
- Window position is automatically adjusted to the center of the work area

## Related Issues
Closes https://github.com/Overload-Technologies/Overload/issues/390

### DIFF
--- a/Sources/Overload/OvGame/src/OvGame/Core/Context.cpp
+++ b/Sources/Overload/OvGame/src/OvGame/Core/Context.cpp
@@ -14,6 +14,35 @@
 using namespace OvCore::Global;
 using namespace OvCore::ResourceManagement;
 
+std::array<int, 4> CalculateOptimalWindowSizeAndPosition(
+	const std::array<int, 4>& p_workAreaSize,
+	std::array<int, 2> p_resolution
+)
+{
+	const auto [workAreaX, workAreaY, workAreaWidth, workAreaHeight] = p_workAreaSize;
+
+	std::array<int, 2> finalResolution = p_resolution;
+
+	// Check if the given resolution fits in the workArea. If not, keep reducing it until it fits
+	while (finalResolution[0] > workAreaWidth || finalResolution[1] > workAreaHeight)
+	{
+		finalResolution[0] /= 2;
+		finalResolution[1] /= 2;
+	}
+
+	if (finalResolution != p_resolution)
+	{
+		OVLOG_WARNING("The target resolution didn't fit in the work area. It has been reduced to " + std::to_string(p_resolution[0]) + "x" + std::to_string(p_resolution[1]));
+	}
+
+	return { 
+		workAreaX + workAreaWidth / 2 - p_resolution[0] / 2,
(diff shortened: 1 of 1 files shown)


Sources/Overload/OvGame/src/OvGame/Core/Context.cpp
```diff
@@ -14,6 +14,35 @@
 using namespace OvCore::Global;
 using namespace OvCore::ResourceManagement;
 
+std::array<int, 4> CalculateOptimalWindowSizeAndPosition(
+	const std::array<int, 4>& p_workAreaSize,
+	std::array<int, 2> p_resolution
```
litelawliet, line 19: We can use the C++20 here by using `std::span`, which overlook the parameter type as long as it is contiguous in memory, just like string_view accepts `char*` and `std::string`, `std::wstring` etc.
In here we can use array, vector or other kind of container that meet the contiguous memory constraint.

Looking then like this (to format correctly):
```C++
std::array<int, 4> CalculateOptimalWindowSizeAndPosition(
	std::span<const int> p_workAreaSize,
	std::span<int> p_resolution)
{
    if (p_workAreaSize.size() < 4)
    {
        OVLOG_ERROR("Invalid work area size. Expected a 4 elements array");
        throw std::invalid_argument("Invalid work area size. Expected a 4 elements array");
    }
    if (p_resolution.size() < 2)
    {
        OVLOG_ERROR("Invalid resolution. Expected a 2 elements array");
		throw std::invalid_argument("Invalid resolution. Expected a 2 elements array");
    }

	const auto& workAreaX = p_workAreaSize[0], &workAreaY = p_workAreaSize[1], &workAreaWidth = p_workAreaSize[2], &workAreaHeight = p_workAreaSize[3];
	std::span<int> finalResolution = p_resolution;

	// Check if the given resolution fits in the workArea. If not, keep reducing it until it fits
	while (finalResolution[0] > workAreaWidth || finalResolution[1] > workAreaHeight)
	{
		finalResolution[0] /= 2;
		finalResolution[1] /= 2;
	}

	if (finalResolution[0] != p_resolution[0] || finalResolution[1] != p_resolution[1])
	{
		OVLOG_WARNING("The target resolution didn't fit in the work area. It has been reduced to " + std::to_string(p_resolution[0]) + "x" + std::to_string(p_resolution[1]));
	}

	return { 
		workAreaX + workAreaWidth / 2 - p_resolution[0] / 2,
		workAreaY + workAreaHeight / 2 - p_resolution[1] / 2,
		p_resolution[0],
		p_resolution[1]
	};
}
```

adriengivry: I like the idea of pushing modern C++ features, however I don't see a case where we would need to support other collection types than `std::array`, with variable parameter counts

litelawliet (approved):

PR #392, opened 2025-02-20
Fixed file rename caused crash in protected folder

adriengivry:
## Description
Removed the need for a temporary file when renaming assets, which caused a crash before when `OvEditor.exe` was located in a protected folder.

## Related Issues
Closes https://github.com/Overload-Technologies/Overload/issues/386

### DIFF
--- a/Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
@@ -944,36 +944,84 @@ void OvEditor::Core::EditorActions::PropagateFileRename(std::string p_previousNa
 	EDITOR_PANEL(Panels::MaterialEditor, "Material Editor").Refresh();
 }
 
-void OvEditor::Core::EditorActions::PropagateFileRenameThroughSavedFilesOfType(const std::string& p_previousName, const std::string& p_newName, OvTools::Utils::PathParser::EFileType p_fileType)
+uint64_t ReplaceStringInFile(const std::filesystem::path& p_filePath,
+	const std::string& p_searchStr,
+	const std::string& p_replaceStr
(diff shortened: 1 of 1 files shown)


Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -970,8 +972,8 @@ void OvEditor::Core::EditorActions::PropagateFileRenameThroughSavedFilesOfType(c
 				out.close(); in.close();
 			}
 
-			std::filesystem::copy_file("TEMP", entry.path(), std::filesystem::copy_options::overwrite_existing);
-			std::filesystem::remove("TEMP");
+			std::filesystem::copy_file(tempFilePath.string(), entry.path(), std::filesystem::copy_options::overwrite_existing);
+			std::filesystem::remove(tempFilePath.string());
```
litelawliet, line 976: I think we could use `c_str()` instead of `string()` whenever possible to prevent multiple allocation.
We're limited by `PathParser::GetFileType()` and `OVLOG_INFO`  which expects `std::string` down the line.

Calling `.string()` will call `_Convert_wide_to` which then call `_Convert_wide_to_narrow` for the char underlying type, which makes allocation. Unless we need the features embedded in the std::string type, we should try to call `c_str()` or at very least get the real type in `std::filesystem`, giving a const reference to _Text which is a `std::wstring`, by calling `native()` instead of `string()`.

Same applies in `ifstream in` and `ofstream out` above.

adriengivry: Updated the implementation, removed unnecessary `string()` calls and also removed the need for a temp file.

litelawliet (approved):

Issue #373, opened 2025-02-15
Expose `OVERLOAD_VERSION` to the whole solution


adriengivry:
**Problem this feature should fix**
The new `OVERLOAD_VERSION` is only visible in the editor

**Expected solution**
Expose `OVERLOAD_VERSION` in the main `premake5.lua` file so it's visible to all the projects.

sameep-git:
Would this just involve moving this line to a higher scope? https://github.com/adriengivry/Overload/blob/d70a7916106bd73ef24e4395f8c134954247f266/Sources/Overload/OvEditor/premake5.lua#L33

adriengivry:
Yep! We would need this in the solution scope, next to the "LUA_SCRIPTING" one

PR #396, opened 2025-02-27
Exposed OVERLOAD_VERSION to entire soltion instead of OvEditor

sameep-git:
Closes: #373 
1. Moved `readVersion()` to `Sources/Overload/premake5.lua` and added `OVERLOAD_VERSION` to `Overload`.
2. Removed `readVersion()` from `Sources/Overload/OvEditor/premake5.lua` and removed `OVERLOAD_VERSION` from `OvEditor` project.

### DIFF
--- a/Sources/Overload/premake5.lua
+++ b/Sources/Overload/premake5.lua
@@ -16,8 +16,7 @@ workspace "Overload"
 	configurations { "Debug", "Release" }
 	platforms { "x64" }
 	startproject "OvEditor"
-	defines { "LUA_SCRIPTING" }
-	defines { "OVERLOAD_VERSION=\"" .. version .. "\"" }
+	defines { "LUA_SCRIPTING", "OVERLOAD_VERSION=\"" .. version .. "\"" }
 
 outputdir = "%{wks.location}/../../Bin/"
 objoutdir = "%{wks.location}/../../Bin-Int/"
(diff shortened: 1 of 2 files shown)


Sources/Overload/premake5.lua
```diff
@@ -1,8 +1,22 @@
+-- Function to read the version number from VERSION.txt
+local function readVersion()
+    local versionFile = io.open("../../VERSION.txt", "r")
```
adriengivry, line 3: Did you test this? I would assume `../../` would bring you outside of the repository, since this `premake5.lua` is one folder up in the hierarchy than `OvEditor/premake5.lua`.

sameep-git: Yes, I tested it! Because the directory is Sources/Overload, we would need `../../` to go to the root directory I think :)

adriengivry: Ah that make sense, so both premake5 share the same workdir (the main premake's path).
> I think :)
Make sure to double check 😝. Otherwise looks good!

sameep-git: I removed one `../` from the premake in OvEditor! I forgot to do it when I was writing the code, and I was not able to generate projects then, but I got it figured out after :)

adriengivry (approved):

Issue #381, opened 2025-02-18
Option for UI elements to be rendered in a separate render pass


adriengivry:
**Problem this feature should fix**
Given that Overload doesn't have any custom components for UI, any UI element (i.e. texts in [Cargo](https://github.com/maxbrundev/Cargo-Demo)), will be rendered in the same render pass as the opaque (or transparent) entities. This leads to post-processing effects being applied after the UI is drawn.

**Expected solution**
A `Material Renderer` should be able to select which render pass this entity should be part of (i.e.: Opaque/Transparent, or UI)

![Image](https://github.com/user-attachments/assets/fe88067b-1123-4ce4-a8e3-9cda7f0a22ef)
_Post-processing effects applied after the UI elements)_

PR #408, opened 2025-03-28
Implemented UI RenderPass

joscanper:
Addresses #381 

* A new UIRenderPass has been added to the SceneRenderer, happening after the PostProcess pass
* I've implemented this option in the material itself (as opposed to the MaterialRenderer) ~~borrowing the concept of Material Domain from UE, in case the PostProcess domain is needed in the future. But both approaches are probably valid, happy to change it if that makes more sense.~~

## Demo
UIRenderPass


Sources/Overload/OvCore/src/OvCore/Rendering/SceneRenderer.cpp
```diff
@@ -265,14 +288,26 @@ OvCore::Rendering::SceneRenderer::AllDrawables OvCore::Rendering::SceneRenderer:
...
 								{
-									transparents.emplace(distanceToActor, drawable);
+									ui.emplace(distanceToActor, drawable);
 								}
 								else
 								{
-									opaques.emplace(distanceToActor, drawable);
+									if (material->IsBlendable())
+									{
+
```
adriengivry, line 300: Unnecessary line break.
There is another one below too.

Sources/Overload/OvCore/src/OvCore/Resources/Material.cpp
```diff
@@ -17,7 +18,7 @@ void OvCore::Resources::Material::OnSerialize(tinyxml2::XMLDocument & p_doc, tin
 	tinyxml2::XMLNode* settingsNode = p_doc.NewElement("settings");
 	p_node->InsertEndChild(settingsNode);
 
-	Serializer::SerializeBoolean(p_doc, settingsNode, "blendable", m_blendable);
```
adriengivry, line 20: I think we should keep that, otherwise we won't be serializing `m_blendable` anymore

Sources/Overload/OvCore/src/OvCore/Resources/Material.cpp
```diff
@@ -5,6 +5,7 @@
 */
 
 #include "OvCore/Resources/Material.h"
+#include "OvRendering/Settings/EMaterialDomain.h"
```
adriengivry, line 8: Let's use `<` `>`, as we are trying to get rid of all `"` includes at some point.
Same for the other includes that this PR adds

Sources/Overload/OvRendering/include/OvRendering/Data/PipelineState.h
```diff
@@ -55,8 +56,9 @@ namespace OvRendering::Data
 				Settings::ERasterizationMode rasterizationMode : 2;
 
 				// B5
-				uint8_t lineWidthPow2 : 3;
-				Settings::EComparaisonAlgorithm depthFunc : 3;
+				uint8_t lineWidthPow2 : 2;
+				Settings::EComparaisonAlgorithm depthFunc : 2;
+				Settings::EMaterialDomain domain : 2;
```
adriengivry, line 61: Since we only have 2 domains, we can use a single bit. Maybe the enum type should be updated too?
I'm woried about changing the `lineWidthPow2` from 3 bits to 2 bits, I don't think it gives us enough precision given the use cases of the `lineWidthPow2` value.

Sources/Overload/OvRendering/include/OvRendering/Settings/EMaterialDomain.h
```diff
@@ -0,0 +1,21 @@
...
+#include <cstdint>
+
+namespace OvRendering::Settings
+{
+	/**
+	* Material Domains
+	*/
+	enum class EMaterialDomain : uint8_t
+	{
+		SURFACE,
```
adriengivry, line 18: Not completely sold on the enum value name. I'm assuming UE uses `SURFACE`, but I'm wondering if we could find something more descriptive. Any other suggestions?

adriengivry (commented):
Great work! There are some minor changes that we could address, but overall it's super cool 🎉

Sources/Overload/OvCore/src/OvCore/Resources/Material.cpp
joscanper: Ooops, good catch 😅

Sources/Overload/OvRendering/include/OvRendering/Settings/EMaterialDomain.h
joscanper: Yeah, I agree, here are some alternatives: WORLD, SOLID, DEFAULT, PRIMARY, BASE, SCENE

Sources/Overload/OvRendering/include/OvRendering/Settings/EMaterialDomain.h
```diff
@@ -0,0 +1,21 @@
...
+*/
+
+#pragma once
+
+#include <cstdint>
+
+namespace OvRendering::Settings
+{
+	/**
+	* Material Domains
```
adriengivry, line 14: We should add a note somewhere where that we can only support 2 values given the pipeline state layout (single bit for domain)

Sources/Overload/OvRendering/include/OvRendering/Settings/EMaterialDomain.h
adriengivry: I like WORLD and SCENE. I'll let you choose 😄

maxbrundev: SURFACE make sense to me, WORLD and SCENE could be interpreted ambiguously since UI is also part of the world/scene, SPATIAL can be good since it clearly distinguishes 3D content from UI elements

joscanper: SPATIAL can also be misinterpreted since UI can be equally placed spatially in the 3D world? 😄 is a tricky one!  Another option would be GEOMETRY (although UI can be rendered on any kind of geometry too). I'm ok with any of these options

maxbrundev: In this case DEFAULT and USER_INTERFACE ?

litelawliet: WORLD makes-sense to me since this means we can have UI within the 3D space, contrary to User-Interface which is expected to be over-the-top

adriengivry: So the scene could be seen as:
- World Pass (Opaque + Transparent)
- Post-Process Pass
- UI Pass (Depth buffer not preserved at this point, so it cannot be world UI)

At some point we could add proper support for a UI render pass that uses the scene depth's buffer, so we could render world-space UI without post-processing effects. Right now the only way to do world-space UI would be with the non-USER_INTERFACE domain, which means that bloom (for instance) will still apply

adriengivry: Alright crazy idea, a boolean called "User Interface". We don't have enough bits in the pipeline state to store more than 2 values anyway 😄

adriengivry: Another approach, which is what Unity is doing, is to expose an integer (render pass) that materials can use to order elements into different passes (essentially exposing the `ERenderPassOrder`). But we can't fit a `uint32_t` into the pipeline state, and we would have to change the way we handle pass to create them dynamically based on the provided render pass values by scene's materials

maxbrundev (approved):

PR #410, opened 2025-03-28
Fixed User Interface materials ignore depth test when gizmo is visible

adriengivry:
## Description
Moved debug render passes after UI, instead of after post-processing.

This fixes the fact that the gizmo is clearing the depth buffer. Another approach would be to use another framebuffer, but this is a lower cost solution.


adriengivry:
Not sure if it's a regression, but I notice the editor UI sometime blinks.

https://github.com/user-attachments/assets/86bfdd6e-7505-4c4e-a2bf-ed7a3fd44db4

litelawliet (approved):

PR #414, opened 2025-03-29
[WIP] Added an option to open a folder in VSCode

### LINKED ISSUE #389: Add a button to the menu bar to open the project in VS code
It's not that convenient to open the file explorer, navigate to the project's folder, and open it with vs code.

**Expected solution**
It would be nice if we could open a project in VS code directly from the editor.

adriengivry:
## Description
* Added an option to open a folder in an external tool (Visual Studio Code or anything else!).
* Instead of adding a button to the menu bar (suggested by #389), I added an option to any folder's contextual menu.
* Added `EditorSettings` entries for custom external tool name and command
* Added `ExternalTools` utility to define the presets
* Added setting to "Menu Bar" to select an external tool preset, or use a custom one.

## Related Issues
Closes #389 

## Screenshot
image

https://github.com/user-attachments/assets/993d7d6e-12a7-42e8-abcc-a3bf81568321


litelawliet (changes requested):
In Release, when opening a folder in VS Code (example), a command-line terminal will open in order to launch the external tool. Keeping the external tool open and opening again a folder from Overload will not create another command-line terminal.
However if you close the external tool prior to open again another folder then a command-line terminal will be created again.

This doesn't happen in Debug because a terminal is attached to Overload, but since no terminal is attached in Release it then have to create one in order to run the command-line that open an external tool.

I think we should hide (set to background the terminal) when running the command line for an external tool.

adriengivry:
> In Release, when opening a folder in VS Code (example), a command-line terminal will open in order to launch the external tool. Keeping the external tool open and opening again a folder from Overload will not create another command-line terminal. However if you close the external tool prior to open again another folder then a command-line terminal will be created again.
> 
> This doesn't happen in Debug because a terminal is attached to Overload, but since no terminal is attached in Release it then have to create one in order to run the command-line that open an external tool.
> 
> I think we should hide (set to background the terminal) when running the command line for an external tool.

Fixed! ✅ https://github.com/Overload-Technologies/Overload/commit/086f45abd44635e134495a72ba7434bd6228fcdf

adriengivry:
Putting this one back to draft, it's currently not possible to know, using `cmd.exe /c` if the underlying command succeeded (`code`, `subl`, or `atom`), therefore, the current solution isn't satisfactory

PR #418, opened 2025-03-30
[WIP] Feature/move scripts folder to assets

### LINKED ISSUE #385: Move scripts to the `Assets/` folder
Scripts are separated into their own `Scripts` folder. However, since scripts are technically assets too, they should also be in the `Assets/` folder.

**Expected solution**
The hierarchy should support script-related operations (create, rename, delete...) directly in the `Assets/` folder, and behaviours should be able to work with scripts located anywhere in the project (currently, scripts in the `Scripts/` folder have no hierarchy).

adriengivry:
## Description
* Removed the requirement to have all the scripts inside of the `Scripts/` folder
* Added folder support for scripts paths
* Implemented a migration tool to automatically move scripts from `Scripts/` to `Assets/Scripts/`, and update all scenes to use the new path

## Related Issues
Fixes https://github.com/Overload-Technologies/Overload/issues/385

## To-Do
- [x] Bug after exiting play mode: scripts become invalid
- [x] Fix rename/move bug
- [ ] Get rid of `projectScriptsPath`
- [ ] Prevent same script from being added multiple times
- [ ] Base the `m_scriptName` on the lua table name instead of the `m_scriptPath.stem()`
- [ ] `TScript` should expose a `GetTypeName()` function
- [ ] Remove the notion of "Behaviour" from the `TScriptEngine`. The only thing `TScriptEngine` should do, is manage the context, create scripts from a path, register/unregister a script to the context, and provide a way to call a function (script, funcName, params)

## Screenshots
image
*When loading a project that uses the old `Scripts/` folder*

image
*Migration tool output (moving scripts from `Scripts/` to `Assets/Scripts/`)*

image
*Asset Browser with no more `Scripts/` folder at the root, but one inside of `Assets/`*

### DIFF
--- a/Sources/Overload/OvCore/include/OvCore/ECS/Actor.h
+++ b/Sources/Overload/OvCore/include/OvCore/ECS/Actor.h
@@ -273,9 +273,9 @@ namespace OvCore::ECS
 
 		/**
 		* Add a behaviour to the actor
-		* @param p_name
+		* @param p_scriptPath
 		*/
-		Components::Behaviour& AddBehaviour(const std::string& p_name);
+		Components::Behaviour& AddBehaviour(const std::string& p_scriptPath);
 
 		/**
 		* Remove a behaviour by refering to the given instance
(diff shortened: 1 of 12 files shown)


adriengivry:
Closed in favor of https://github.com/Overload-Technologies/Overload/pull/669

PR #420, opened 2025-03-30
Removed fake root actor

adriengivry:
## Description
Removed the fake "Root" actor in the editor's hierarchy, which was only there because panels (windows) didn't support drag-and-drop, so we had to have this fake root actor to provide a way to detach an actor from its parent.

## Screenshots
image

https://github.com/user-attachments/assets/888509f9-de19-4792-a78e-569becb6821e

### DIFF
--- a/Sources/Overload/OvEditor/include/OvEditor/Panels/Hierarchy.h
+++ b/Sources/Overload/OvEditor/include/OvEditor/Panels/Hierarchy.h
@@ -84,8 +84,6 @@ namespace OvEditor::Panels
 		OvTools::Eventing::Event<OvCore::ECS::Actor&> ActorUnselectedEvent;
 
 	private:
-		OvUI::Widgets::Layout::TreeNode* m_sceneRoot;
-
 		std::unordered_map<OvCore::ECS::Actor*, OvUI::Widgets::Layout::TreeNode*> m_widgetActorLink;
 	};
 }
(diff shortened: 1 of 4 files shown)


adriengivry:
Reported crash:
```
 	OvEditor.exe!OvUI::Internal::WidgetContainer::DrawWidgets() Line 88	C++
>	OvEditor.exe!OvUI::Panels::PanelWindow::_Draw_Impl() Line 177	C++
 	OvEditor.exe!OvUI::Panels::APanel::Draw() Line 22	C++
 	OvEditor.exe!OvUI::Modules::Canvas::Draw() Line 39	C++
 	OvEditor.exe!OvUI::Core::UIManager::Render() Line 283	C++
 	OvEditor.exe!OvEditor::Core::Editor::RenderEditorUI(float p_deltaTime) Line 242	C++
 	OvEditor.exe!OvEditor::Core::Editor::Update(float p_deltaTime) Line 92	C++
 	OvEditor.exe!OvEditor::Core::Application::Run() Line 28	C++
 	OvEditor.exe!TryRun(const std::string & projectPath, const std::string & projectName) Line 103	C++
 	OvEditor.exe!main(int argc, char * * argv) Line 78	C++
```

litelawliet (approved):
The bug mentionned in this PR (https://github.com/Overload-Technologies/Overload/pull/420#issuecomment-2764777913) was indeed not part of the PR. But since it falls in the same topic it is welcome to have a fix from here.

PR #431, opened 2025-04-01
Editor Layouts Management

maxbrundev:
## Description
Added Editor Layouts management

## Screenshots
layoutstest

### DIFF
--- a/Sources/Overload/OvEditor/src/OvEditor/Core/Context.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Core/Context.cpp
@@ -154,6 +154,7 @@ OvEditor::Core::Context::Context(const std::string& p_projectPath, const std::st
 	ServiceLocator::Provide<OvAudio::Core::AudioPlayer>(*audioPlayer);
 	ServiceLocator::Provide<OvCore::Scripting::ScriptEngine>(*scriptEngine);
 	ServiceLocator::Provide<OvEditor::Utils::TextureRegistry>(*textureRegistry);
+	ServiceLocator::Provide<OvUI::Core::UIManager>(*uiManager);
 
 	ApplyProjectSettings();
 }
--- a/Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
@@ -377,6 +377,21 @@ void OvEditor::Core::EditorActions::ResetLayout()
     DelayAction([this]() {m_context.uiManager->ResetLayout("Config\\layout.ini"); });
 }
 
+void OvEditor::Core::EditorActions::SaveLayout(const std::string& p_fileName)
+{
+	DelayAction([&]() {m_context.uiManager->SaveLayout(std::ref(p_fileName)); });
+}
+
+void OvEditor::Core::EditorActions::SaveCurrentLayout()
+{
+	DelayAction([&]() {m_context.uiManager->SaveCurrentLayout(); });
+}
+
+void OvEditor::Core::EditorActions::SetLayout(const std::string& p_fileName)
+{
+	DelayAction([&]() {m_context.uiManager->SetLayout(std::ref(p_fileName)); });
+}
+
 void OvEditor::Core::EditorActions::SetSceneViewCameraSpeed(int p_speed)
 {
 	EDITOR_PANEL(Panels::SceneView, "Scene View").GetCameraController().SetSpeed((float)p_speed);
--- a/Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
@@ -4,6 +4,8 @@
 * @licence: MIT
 */
 
+#include <filesystem>
+
 #include <OvTools/Utils/SystemCalls.h>
 
 #include <OvCore/ECS/Components/CCamera.h>
@@ -30,6 +32,7 @@
 #include "OvEditor/Core/EditorActions.h"
 #include "OvEditor/Settings/EditorSettings.h"
 #include "OvEditor/Utils/ActorCreationMenu.h"
+#include "OvUI/Plugins/ContextualMenu.h"
 
 using namespace OvUI::Panels;
 using namespace OvUI::Widgets;
@@ -197,8 +200,90 @@ void OvEditor::Panels::MenuBar::CreateSettingsMenu()
 
 void OvEditor::Panels::MenuBar::CreateLayoutMenu() 
 {
-	auto& layoutMenu = CreateWidget<MenuList>("Layout");
-	layoutMenu.CreateWidget<MenuItem>("Reset").ClickedEvent += EDITOR_BIND(ResetLayout);
+	auto& layoutMenuList = CreateWidget<MenuList>("Layout");
+
+	auto& saveMenuItem = layoutMenuList.CreateWidget<MenuItem>("Save");
+	saveMenuItem.ClickedEvent += EDITOR_BIND(SaveCurrentLayout);
+
+	auto& saveNewMenuList = layoutMenuList.CreateWidget<MenuList>("Save New");
+	auto& layoutInputText = saveNewMenuList.CreateWidget<InputFields::InputText>("Layout Name");
+	layoutInputText.selectAllOnClick = true;
+	layoutInputText.EnterPressedEvent += [this](std::string p_input)
+	{
+		if (p_input.empty())
+			return;
+
+		auto& UIManager = OvCore::Global::ServiceLocator::Get<OvUI::Core::UIManager>();
+		std::string layoutsPath = UIManager.GetLayoutsPath();
+		EDITOR_EXEC(DelayAction(std::bind(&OvUI::Core::UIManager::SaveLayout, &UIManager, layoutsPath + p_input + ".ini"), 1));
+	};
+
+	auto& loadMenuList = layoutMenuList.CreateWidget<MenuList>("Load");
+
+	loadMenuList.ClickedEvent += [&]
+	{
+		loadMenuList.RemoveAllWidgets();
+
+		auto& UIManager = OvCore::Global::ServiceLocator::Get<OvUI::Core::UIManager>();
+		std::string layoutsPath = UIManager.GetLayoutsPath();
+
+		for (const auto& entry : std::filesystem::directory_iterator(layoutsPath))
+		{
+			if (entry.is_regular_file() && entry.path().extension() == ".ini")
+			{
+				std::shared_ptr<std::string> layoutFileName = std::make_shared<std::string>(entry.path().filename().string());
+
+				auto& layoutMenuItem = loadMenuList.CreateWidget<MenuItem>(*layoutFileName);
+				layoutMenuItem.name = OvTools::Utils::PathParser::GetFileWithoutExtension(*layoutFileName);
+
+				layoutMenuItem.ClickedEvent += [this, layoutsPath, layoutFileName]
+				{
+					auto& UIManager = OvCore::Global::ServiceLocator::Get<OvUI::Core::UIManager>();
+					std::string layoutsPath = UIManager.GetLayoutsPath();
+					EDITOR_EXEC(DelayAction(std::bind(&OvUI::Core::UIManager::SetLayout, &UIManager, layoutsPath + *layoutFileName), 1));
+				};
+				
+				auto& contextualMenu = layoutMenuItem.AddPlugin<OvUI::Plugins::ContextualMenu>();
+				auto& deleteMenuItem = contextualMenu.CreateWidget<MenuItem>("Delete");
+
+				deleteMenuItem.ClickedEvent += [this, layoutFileName, &layoutMenuItem]
+				{
+					//EDITOR_EXEC(ResetToDefaultLayout);
+					auto& UIManager = OvCore::Global::ServiceLocator::Get<OvUI::Core::UIManager>();
+					std::string layoutsPath = UIManager.GetLayoutsPath();
+					EDITOR_EXEC(DelayAction(std::bind(&OvUI::Core::UIManager::DeleteLayout, &UIManager, layoutsPath + *layoutFileName), 1));
+					layoutMenuItem.enabled = false;
+
+				};
+				auto& renameToMenuList = contextualMenu.CreateWidget<MenuList>("Rename to...");
+
+				auto& renameInputText = renameToMenuList.CreateWidget<InputFields::InputText>("");
+				renameInputText.content = OvTools::Utils::PathParser::GetFileWithoutExtension(*layoutFileName);
+				renameInputText.selectAllOnClick = true;
+
+				renameInputText.EnterPressedEvent += [this, layoutFileName, &contextualMenu, &layoutMenuItem](std::string p_newName)
+				{
+					if (p_newName.empty())
+						return;
+
+					layoutMenuItem.name = p_newName;
+					//EDITOR_EXEC(ResetToDefaultLayout);
+
+					auto& UIManager = OvCore::Global::ServiceLocator::Get<OvUI::Core::UIManager>();
+					std::string layoutsPath = UIManager.GetLayoutsPath();
+
+					std::string oldFileName = layoutsPath + *layoutFileName;
+					std::string newFileName = layoutsPath + p_newName + ".ini";
+					EDITOR_EXEC(DelayAction(std::bind(&OvUI::Core::UIManager::RenameLayout, &UIManager, oldFileName, newFileName), 1));
+
+					*layoutFileName = p_newName + ".ini";
+					contextualMenu.Close();
+				};
+			}
+		}
+	};
+
+	layoutMenuList.CreateWidget<MenuItem>("Reset").ClickedEvent += EDITOR_BIND(ResetLayout);
 }
 
 void OvEditor::Panels::MenuBar::CreateHelpMenu()
--- a/Sources/Overload/OvTools/include/OvTools/Utils/PathParser.h
+++ b/Sources/Overload/OvTools/include/OvTools/Utils/PathParser.h
@@ -67,6 +67,12 @@ namespace OvTools::Utils
 		static std::string GetExtension(const std::string& p_path);
 
 		/**
+		* Returns the file without the extension
+		* @param p_file
+		*/
+		static std::string GetFileWithoutExtension(const std::string& p_file);
+
+		/**
 		* Convert the EFileType value to a string
 		* @param p_fileType
 		*/
--- a/Sources/Overload/OvUI/include/OvUI/Core/UIManager.h
+++ b/Sources/Overload/OvUI/include/OvUI/Core/UIManager.h
@@ -96,11 +96,47 @@ namespace OvUI::Core
 		*/
 		void EnableDocking(bool p_value);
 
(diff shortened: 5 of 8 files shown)


Sources/Overload/OvEditor/src/OvEditor/Core/Context.cpp
```diff
@@ -154,6 +154,7 @@ OvEditor::Core::Context::Context(const std::string& p_projectPath, const std::st
 	ServiceLocator::Provide<OvAudio::Core::AudioPlayer>(*audioPlayer);
 	ServiceLocator::Provide<OvCore::Scripting::ScriptEngine>(*scriptEngine);
 	ServiceLocator::Provide<OvEditor::Utils::TextureRegistry>(*textureRegistry);
+	ServiceLocator::Provide<OvUI::Core::UIManager>(*uiManager);
```
adriengivry, line 157: Wondering if we need to expose the `UIManager`, since you can get it from anywhere in `OvEditor` with `EDITOR_CONTEXT(uiManager)`. Services are mostly for the `OvCore` module to access other systems, without a global context

Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -377,6 +377,21 @@ void OvEditor::Core::EditorActions::ResetLayout()
     DelayAction([this]() {m_context.uiManager->ResetLayout("Config\\layout.ini"); });
 }
 
+void OvEditor::Core::EditorActions::SaveLayout(const std::string& p_fileName)
+{
+	DelayAction([&]() {m_context.uiManager->SaveLayout(std::ref(p_fileName)); });
```
adriengivry, line 382: `std::ref` is mostly to force passing a reference in cases where it's ambiguous (multiple function overloads). In this case it isn't necessary.

Sources/Overload/OvUI/src/OvUI/Core/UIManager.cpp
```diff
@@ -6,7 +6,13 @@
 
 #include "OvUI/Core/UIManager.h"
 
-OvUI::Core::UIManager::UIManager(GLFWwindow* p_glfwWindow, Styling::EStyle p_style, std::string_view p_glslVersion)
+#include <filesystem>
+
+#include "OvTools/Utils/SystemCalls.h"
```
adriengivry, line 11: Let's favour `<` `>` instead of `"` for inclues

Sources/Overload/OvUI/include/OvUI/Core/UIManager.h
```diff
@@ -133,5 +171,7 @@ namespace OvUI::Core
 		Modules::Canvas* m_currentCanvas = nullptr;
 		std::unordered_map<std::string, ImFont*> m_fonts;
 		std::string m_layoutSaveFilename = "imgui.ini";
+		const std::string m_defaultLayout;
+		const std::string m_layoutsPath;
```
adriengivry, line 175: I'm thinking we should start using `std::filesystem::path` when storing path, as it's safer, and allow operations like:
```cpp
std::filesystem::path("path/to/something") / "folder" / "other_folder" / "file.txt"
```

adriengivry (changes requested):
image

image


Looks like it's also displaying all the .ini files I have in `%appdata%/OverloadTech/OvEditor`.

The layouts should probably be stored in their own folder (`Layouts/`)

Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
```diff
@@ -197,8 +200,90 @@ void OvEditor::Panels::MenuBar::CreateSettingsMenu()
...
+		loadMenuList.RemoveAllWidgets();
+
+		auto& UIManager = OvCore::Global::ServiceLocator::Get<OvUI::Core::UIManager>();
+		std::string layoutsPath = UIManager.GetLayoutsPath();
+
+		for (const auto& entry : std::filesystem::directory_iterator(layoutsPath))
+		{
+			if (entry.is_regular_file() && entry.path().extension() == ".ini")
+			{
+				std::shared_ptr<std::string> layoutFileName = std::make_shared<std::string>(entry.path().filename().string());
```
adriengivry, line 234: Any reason why we are allocating this string dynamically, and with a `shared_ptr`? I don't see any other code path that shares the ownership of this memory location

Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
```diff
@@ -197,8 +200,90 @@ void OvEditor::Panels::MenuBar::CreateSettingsMenu()
...
+					EDITOR_EXEC(DelayAction(std::bind(&OvUI::Core::UIManager::SetLayout, &UIManager, layoutsPath + *layoutFileName), 1));
+				};
+				
+				auto& contextualMenu = layoutMenuItem.AddPlugin<OvUI::Plugins::ContextualMenu>();
+				auto& deleteMenuItem = contextualMenu.CreateWidget<MenuItem>("Delete");
+
+				deleteMenuItem.ClickedEvent += [this, layoutFileName, &layoutMenuItem]
+				{
+					//EDITOR_EXEC(ResetToDefaultLayout);
+					auto& UIManager = OvCore::Global::ServiceLocator::Get<OvUI::Core::UIManager>();
```
adriengivry, line 252: Should probably be lower camel case

Sources/Overload/OvTools/include/OvTools/Utils/PathParser.h
```diff
@@ -66,6 +66,12 @@ namespace OvTools::Utils
 		*/
 		static std::string GetExtension(const std::string& p_path);
 
+		/**
+		* Returns the file without the extension
+		* @param p_file
+		*/
+		static std::string GetFileWithoutExtension(const std::string& p_file);
```
adriengivry, line 73: Might not be necessary, since this function would return `/path/to/foo` if given `/path/to/foo.txt`, which is fairly uncommon.

Alternatively, `std::filesystem::path::stem()` would return `foo`, and `std::filesystem::path::filename()` would return `foo.txt`.

So maybe in this use case, `stem()` would be the way to go

Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
```diff
@@ -197,8 +200,90 @@ void OvEditor::Panels::MenuBar::CreateSettingsMenu()
...
+					//EDITOR_EXEC(ResetToDefaultLayout);
+
+					auto& UIManager = OvCore::Global::ServiceLocator::Get<OvUI::Core::UIManager>();
+					std::string layoutsPath = UIManager.GetLayoutsPath();
+
+					std::string oldFileName = layoutsPath + *layoutFileName;
+					std::string newFileName = layoutsPath + p_newName + ".ini";
+					EDITOR_EXEC(DelayAction(std::bind(&OvUI::Core::UIManager::RenameLayout, &UIManager, oldFileName, newFileName), 1));
+
+					*layoutFileName = p_newName + ".ini";
```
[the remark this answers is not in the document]
maxbrundev: Old shenanigans don't even remember what I tried here, removed!

Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
```diff
@@ -202,31 +202,30 @@ void OvEditor::Panels::MenuBar::CreateLayoutMenu()
...
-		std::string layoutsPath = UIManager.GetLayoutsPath();
-		EDITOR_EXEC(DelayAction(std::bind(&OvUI::Core::UIManager::SaveLayout, &UIManager, layoutsPath + p_input + ".ini"), 1));
+		auto& uiManager = *EDITOR_CONTEXT(uiManager);
+		EDITOR_EXEC(DelayAction(std::bind(&OvUI::Core::UIManager::SaveLayout, &uiManager, layoutsPath + p_input + ".ini"), 1));
 	};
 
 	auto& loadMenuList = layoutMenuList.CreateWidget<MenuList>("Load");
 
-	loadMenuList.ClickedEvent += [&]
+	loadMenuList.ClickedEvent += [&, layoutsPath]
```
adriengivry, line 225: You shouldn't need to copy `layoutsPath` if you already capture it by reference with `&`. I'd advise against using `&`, as it makes it hard to catch potential issues, so explicitly using `&layoutsPath` could be a good option. Not sure how many other references you need here

Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
```diff
@@ -261,20 +259,18 @@ void OvEditor::Panels::MenuBar::CreateLayoutMenu()
 				renameInputText.content = OvTools::Utils::PathParser::GetFileWithoutExtension(layoutFileName);
 				renameInputText.selectAllOnClick = true;
 
-				renameInputText.EnterPressedEvent += [this, layoutFileName, &contextualMenu, &layoutMenuItem](std::string p_newName)
+				renameInputText.EnterPressedEvent += [this, layoutsPath, layoutFileName, &contextualMenu, &layoutMenuItem](std::string p_newName)
```
adriengivry, line 262: Same here, might not want to copy `layoutsPath` and `layoutFileName`

Sources/Overload/OvEditor/src/OvEditor/Core/EditorActions.cpp
maxbrundev: Old shenanigans again was thinking a copy was done without it back then, forgot to remove, done!

Sources/Overload/OvUI/src/OvUI/Core/UIManager.cpp
```diff
@@ -17,6 +23,8 @@ OvUI::Core::UIManager::UIManager(GLFWwindow* p_glfwWindow, Styling::EStyle p_sty
 	
 	ImGui_ImplGlfw_InitForOpenGL(p_glfwWindow, true);
 	ImGui_ImplOpenGL3_Init(p_glslVersion.data());
+
+	std::filesystem::create_directory(m_layoutsPath);
```
adriengivry, line 27: Crashes if this gets called when the user starts Overload for the first time (If no `%appdata%/OverloadTech`) folder already exists.
Instead, you can use:
```suggestion
	std::filesystem::create_directories(m_layoutsPath);
```

Sources/Overload/OvUI/src/OvUI/Core/UIManager.cpp
```diff
@@ -6,7 +6,13 @@
 
 #include "OvUI/Core/UIManager.h"
 
-OvUI::Core::UIManager::UIManager(GLFWwindow* p_glfwWindow, Styling::EStyle p_style, std::string_view p_glslVersion)
+#include <filesystem>
+
+#include <OvTools/Utils/SystemCalls.h>
+
+OvUI::Core::UIManager::UIManager(GLFWwindow* p_glfwWindow, Styling::EStyle p_style, std::string_view p_glslVersion) :
+m_defaultLayout("Config\\layout.ini"),
```
adriengivry, line 14:
```suggestion
m_defaultLayout{ std::filesystem::path{} / "Config" / "layout.ini" }
```

Sources/Overload/OvUI/src/OvUI/Core/UIManager.cpp
```diff
@@ -227,11 +235,11 @@ bool OvUI::Core::UIManager::IsEditorLayoutSaveEnabled() const
...
 }
 
-void OvUI::Core::UIManager::SetEditorLayoutSaveFilename(const std::string & p_filename)
+void OvUI::Core::UIManager::SetEditorLayoutSaveFilename(const std::filesystem::path& p_filePath)
 {
-	m_layoutSaveFilename = p_filename;
+	m_layoutSaveFilename = p_filePath;
 	if (IsEditorLayoutSaveEnabled())
-		ImGui::GetIO().IniFilename = m_layoutSaveFilename.c_str();
+		ImGui::GetIO().IniFilename = m_layoutSaveFilename.string().c_str();
```
adriengivry, line 242: This introduces a bug where the editor layout doesn't get saved anymore when quitting.
This is due to the return value of `m_layoutSaveFilename.string().c_str();` becoming a dangling pointer, since the generated `std::string` with `string()` is destroyed on the next line, while ImGui requires this string to stay alive until the editor closes

adriengivry: `ImGui::GetIO().IniFilename` is non owning, so we need to have an allocated string somewhere in memory

Sources/Overload/OvUI/src/OvUI/Core/UIManager.cpp
```diff
@@ -254,11 +262,60 @@ void OvUI::Core::UIManager::EnableDocking(bool p_value)
...
+	SetEditorLayoutSaveFilename(p_filePath);
+
+	ImGui::SaveIniSettingsToDisk(m_layoutSaveFilename.string().c_str());
+}
+
+void OvUI::Core::UIManager::SaveCurrentLayout()
+{
+	if (!std::filesystem::exists(m_layoutSaveFilename))
+	{
+		m_layoutSaveFilename = (m_layoutsPath / "layout.ini").string();
```
adriengivry, line 291: `.string()` here generate string, that is then used to generate another instance of `std::filesystem::path`.
```suggestion
		m_layoutSaveFilename = m_layoutsPath / "layout.ini";
```

Sources/Overload/OvUI/include/OvUI/Core/UIManager.h
```diff
@@ -96,11 +97,53 @@ namespace OvUI::Core
...
+		* Reset the UI layout to the given configuration file
+		* @param p_config
+		*/
+		void ResetLayout(const std::string& p_config) const;
+
+		/**
+		* Load the UI layout from the given configuration file
+		* @param p_fileName
+		*/
+		void LoadLayout(const std::string& p_fileName);
```
adriengivry, line 116: Looks like it's unused

Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
```diff
@@ -197,8 +200,78 @@ void OvEditor::Panels::MenuBar::CreateSettingsMenu()
...
+
+				renameInputText.EnterPressedEvent += [entry, &layoutMenuItem, &layoutsPath, &contextualMenu](std::string p_newName)
+				{
+					if (p_newName.empty())
+						return;
+
+					layoutMenuItem.name = p_newName;
+					
+					auto& uiManager = *EDITOR_CONTEXT(uiManager);
+					EDITOR_EXEC(DelayAction(std::bind(&OvUI::Core::UIManager::RenameLayout, &uiManager, entry, layoutsPath / (p_newName + ".ini")), 1));
```
adriengivry, line 266: Crashes when performed multiple times without closing the containing menu list.

https://github.com/user-attachments/assets/e44f7792-1915-41f6-89d3-3e9bbc269139

adriengivry: Also, seems like the action delay isn't necessary. Got it to work with:
```cpp
uiManager.RenameLayout(entry, layoutsPath / (p_newName + ".ini"));
```

Sources/Overload/OvUI/include/OvUI/Core/UIManager.h
```diff
@@ -132,6 +177,8 @@ namespace OvUI::Core
 		bool m_dockingState;
 		Modules::Canvas* m_currentCanvas = nullptr;
 		std::unordered_map<std::string, ImFont*> m_fonts;
-		std::string m_layoutSaveFilename = "imgui.ini";
+		std::filesystem::path m_layoutSaveFilename = "imgui.ini";
```
adriengivry, line 180: Probably could be made a `std::string` (would solve the imgui ini save bug) since path features are never used with this one. Also I've noticed that it isn't saved, so restarting the editor results in the "current layout" being layout.ini again, and it would need to be loaded manually again.

Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
maxbrundev: Yep we are iterating other the entries resulting in not updated path when we pass the first parameter to RenameLayout

Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
```diff
@@ -237,6 +237,8 @@ void OvEditor::Panels::MenuBar::CreateLayoutMenu()
 				{
 					auto& uiManager = *EDITOR_CONTEXT(uiManager);
 					EDITOR_EXEC(DelayAction(std::bind(&OvUI::Core::UIManager::SetLayout, &uiManager, entry.path()), 1));
+
+					Settings::EditorSettings::LatestLayout = entry.path().string();
```
adriengivry, line 241: We could save just the filename (`.stem()`), so we can retrieve it afterwards at runtime, since there is no hierarchy inside of the `Layouts/` folder.

If the string `LatestLayout` is empty, then load default

Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
maxbrundev: Fixed, I now remember why I was using the `shared_ptr` before

maxbrundev: > Also, seems like the action delay isn't necessary. Got it to work with:
> 
> ```c++
> uiManager.RenameLayout(entry, layoutsPath / (p_newName + ".ini"));
> ```
> 
> Wondering if other actions also need to be delayed (`DeleteLayout`, `SetLayout`...)

`ResetToDefaultLayout` has to be delayed, otherwise, it results in un-docked panels with random positions due to the UI structure changing at the same time it is being rendered. 
I suggest keeping, at minimum, all methods that encapsulate `LoadIniSettingsFromDisk` in a delayed call.

Sources/Overload/OvEditor/src/OvEditor/Panels/MenuBar.cpp
```diff
@@ -216,8 +219,90 @@ void OvEditor::Panels::MenuBar::CreateSettingsMenu()
...
+	loadMenuList.ClickedEvent += [&]
+	{
+		loadMenuList.RemoveAllWidgets();
+
+		for (const auto& entry : std::filesystem::directory_iterator(layoutsPath))
+		{
+			if (entry.is_regular_file() && entry.path().extension() == ".ini")
+			{
+				auto& layoutMenuItem = loadMenuList.CreateWidget<MenuItem>(entry.path().stem().string());
+				layoutMenuItem.name = entry.path().stem().string();
```
adriengivry, line 253: We could display which layout is currently selected with a checkmark
```suggestion
				layoutMenuItem.name = entry.path().stem().string();
				layoutMenuItem.checkable = true;
				layoutMenuItem.checked = Settings::EditorSettings::LatestLayout.Get() == entry.path().stem().string();
```

image

adriengivry: Also, since with this change we can show the currently selected layout, maybe we should always have a "Default" option in the "Load" list, and the "Default" option would have no contextual menu (so it cannot be deleted)

PR #438, opened 2025-04-03
Moved ImGui and tinyxml2 outside of Overload SDK, and added projects for each dependency

adriengivry:
## Description
Moved `imgui` and `tinyxml2` into their own folder under "Dependencies", and added projects for all Overload dependencies.

Updated `Overload` solution premake to segment its projects into 4 sections:
* Dependencies (Prebuilt + built by our solution)
* Overload Applications (Executables)
* Overload SDK (Modules)
* Resources (was already there)

## Notes
Also used this opportunity to alphabetize includes in the affected files.

## Screenshots
image

image


Sources/Overload/premake5.lua
```diff
@@ -24,18 +24,27 @@ dependdir = "%{wks.location}/../../Dependencies/"
...
-include "OvDebug"
-include "OvMaths"
-include "OvPhysics"
-include "OvRendering"
-include "OvTools"
-include "OvUI"
-include "OvWindowing"
+group "Dependencies"
+    include "../ImGui"
+    include "../tinyxml2"
```
adriengivry, line 29: Wondering if header only libraries (like sol2), should be added there too? I don't think so, since they cannot technically be compiled. I guess this new "Dependencies" folder should only have sources that can be compiled.

litelawliet (approved):

PR #440, opened 2025-04-04
Editor lag on many logs

### LINKED ISSUE #314: The longer the log, the greater the lag.
**Description**
Adding logs to the console make the whole editor lag.

**To Reproduce**
1. Log something in the console every frame (i.e. break a shader)
2. Notice the issue

**Expected behavior**
We should stop displaying old logs if there are too many.

**Screenshots**
N/A

Synyproxy:
CLOSES #314

Having many messages in the console causes the editor to have frame drops.
Limited the amount of messages to a value that defaults at 500.
Users can change it in the menu bar -> Settings -> Console Settings -> Max Logs

Could serialize this data so user choices persists.

### DIFF
--- a/Sources/Overload/OvEditor/include/OvEditor/Settings/EditorSettings.h
+++ b/Sources/Overload/OvEditor/include/OvEditor/Settings/EditorSettings.h
@@ -94,5 +94,6 @@ namespace OvEditor::Settings
 		inline static Property<float> RotationSnapUnit = { 15.0f };
 		inline static Property<float> ScalingSnapUnit = { 1.0f };
 		inline static Property<int> ColorTheme = { static_cast<int>(OvUI::Styling::EStyle::DEFAULT_DARK) };
+		inline static Property<int> ConsoleMaxLogs = { 500 };
 	};
 }
--- a/Sources/Overload/OvEditor/src/OvEditor/Panels/Console.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Panels/Console.cpp
@@ -13,6 +13,7 @@
 #include <OvUI/Widgets/Selection/CheckBox.h>
 #include <OvUI/Widgets/Visual/Separator.h>
 #include <OvUI/Widgets/Layout/Spacing.h>
+#include "OvEditor/Settings/EditorSettings.h"
 
 using namespace OvUI::Panels;
 using namespace OvUI::Widgets;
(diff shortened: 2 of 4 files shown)


Sources/Overload/OvEditor/include/OvEditor/Settings/EditorSettings.h
```diff
@@ -94,5 +94,6 @@ namespace OvEditor::Settings
 		inline static Property<float> RotationSnapUnit = { 15.0f };
 		inline static Property<float> ScalingSnapUnit = { 1.0f };
 		inline static Property<int> ColorTheme = { static_cast<int>(OvUI::Styling::EStyle::DEFAULT_DARK) };
+		inline static Property<int> ConsoleMaxLogs = { 500 };
```
adriengivry, line 97: Anything added here ideally gets serialized in EditorSettings.cpp

Sources/Overload/OvEditor/src/OvEditor/Panels/Console.cpp
```diff
@@ -13,6 +13,7 @@
 #include <OvUI/Widgets/Selection/CheckBox.h>
 #include <OvUI/Widgets/Visual/Separator.h>
 #include <OvUI/Widgets/Layout/Spacing.h>
+#include "OvEditor/Settings/EditorSettings.h"
```
adriengivry, line 16: I know I still haven't updated to coding conventions, but moving forward, includes will be sorted in alphabetical order (lower case first) and use <> instead of ""

Sources/Overload/OvEditor/src/OvEditor/Panels/Console.cpp
```diff
@@ -100,6 +101,11 @@ void OvEditor::Panels::Console::OnLogIntercepted(const OvDebug::LogData & p_logD
 	consoleItem1.enabled = IsAllowedByFilter(p_logData.logLevel);
 
 	m_logTextWidgets[&consoleItem1] = p_logData.logLevel;
+
+	if (m_logGroup->GetWidgets().size() > Settings::EditorSettings::ConsoleMaxLogs)
+	{
+		m_logGroup->RemoveWidget(*m_logGroup->GetWidgets().front().first);
```
adriengivry, line 107: What happens if there are already 100 logs in the console, and the user set the max to 10? I feel like the user will always have at least as many logs as he already had before he set the limit

PR #446, opened 2025-04-10
Updated ImGui to v1.91.9

adriengivry:
## Description
Updated ImGui to [v1.91.9-docking](https://github.com/ocornut/imgui/releases/tag/v1.91.9-docking).

The goal of this update is also to keep `ImGui` as "vanilla" as possible, instead of modifying its source code like we did in the past.

Also updated the `InputManager` to expose `GetMouseScroll`, since [`ImGui::GetScroll` function is broken(returns `0.0f`)](https://github.com/ocornut/imgui/issues/7546).
* Exposed the new `GetMouseScroll` method to Lua
* Updated the lua scripting API ([Inputs](https://github.com/Overload-Technologies/Overload/wiki/Inputs)) to reflect the changes

## Related Issues
Fixes https://github.com/Overload-Technologies/Overload/issues/121

## To-Do
- [x] Address missing `ImGui::GetIO().DisableMouseUpdate` (was added by us to ImGui). Wondering if there is a better way to handle that without modifying ImGui's sources
- [x] Mouse scroll doesn't work anymore
- [x] ~~Address `forceHover` not used with plots (`PlotHistogram` and `PlotLines`). I recall it was also custom code added to ImGui, can we do without it?~~ follow-up issue: https://github.com/Overload-Technologies/Overload/issues/448
- [x] Update `layout.ini`


litelawliet (approved):

litelawliet:
Just to be sure, is it version 1.9191 or is it 1.91.9b ?

PR #449, opened 2025-04-12
Bump lua 5.4.7 & sol 3.5.0

litelawliet:
Here's a list of changes from Lua 5.3 to current 5.4.7:
https://www.lua.org/manual/5.4/readme.html#changes

Lua is now a static lib and is named lua.lib, changes applies in OvEditor and OvGame's premake5 files.
Lua and sol are now separated project in Dependencies.
Sol also got updated to latest version available on GitHub (3.5.0—to get fixes for Lua 5.4), as of now : https://github.com/ThePhD/sol2/commit/c1f95a773c6f8f4fde8ca3efe872e7286afe4444

In cargo:
Currently having an issue getting the light color from "Key Light" in FinalGateEnigma, making impossible to finish the final enigma. We're setting the light color of Key Light in OnStart, we should have it set in OnAwake instead. We're trying to access the light color in FinalGateEnigma in OnStart function, thus leading to Undefined Behavior. The fact it was working previously is pure luck. Bumping Lua and Sol reavealed this UB due to getting the light color and setting the light color at the same time in OnStart functions, no ordering was guaranteed.

Cargo's fix https://github.com/Overload-Technologies/Cargo-Demo/pull/13.


Sources/Overload/OvGame/premake5.lua
```diff
@@ -35,7 +35,8 @@ project "OvGame"
 
 	filter { "configurations:Release" }
 		defines { "NDEBUG" }
-		optimize "On"
+		optimize "Speed"
+		flags { "LinkTimeOptimization" }
```
adriengivry, line 39: Deprecated in Premake 5.0.0-beta4. Use `linktimeoptimization` API instead.

adriengivry (approved):
Approved conditionally that `lua.c` and `onelua.c` files are removed, as well as the deprecated flag is replaced by [`linktimeoptimization “On”`](https://premake.github.io/docs/linktimeoptimization).

Sources/Overload/OvGame/premake5.lua
litelawliet: Since we have an older version of premake than the one you're mentionning, the setting "linkoptimization 'value'" isn't available yet.
We'll need to make a bump of premake and modify our premake5.lua files if needed to reflect the change.

PR #470, opened 2025-04-30
Exposed shader feature set to the material editor

adriengivry:
## Description
- Updated material to store a "current feature set"
- Exposed the current feature set through the material editor
- Updated material binding to use the program matching the current feature set
- Added serialization/deserialization of feature set
- Added an option to override the feature set of a material (for render passes to enforce some features, i.e. depth only, debug...)
- Updated material editor to show all uniforms, even the ones that are feature-specific

_Follow-up work: we should update the material editor so that widgets for shader uniforms and features are updated after a shader has been recompiled. Right now, only unselecting/reselecting the material, or changing the shader will trigger uniforms/features to be updated to reflect the last compilation result._

## Review Guidance
_N/A_

## To-Do
- [x] Make sure that if some uniforms are inside a `#ifdef` block, they are still displayed in the editor
- [x] Handle case when the same uniform name is used by 2 variants for different types/purposes

## Screenshots

https://github.com/user-attachments/assets/2cd860a2-b382-4367-9b61-8c10de271db0


Synyproxy (approved):
Very Cool, Works as intended. I have tested creating a shader and adding a variant

PR #477, opened 2025-05-01
Fixed Asset View when editing material (#474)

### LINKED ISSUE #474: Asset view not gaining focus when editing a material
**Description**
Double-clicking on a material, or right-clicking on it and selecting "Edit" opens the material in the material editor, but doesn't focus the asset view.

**To Reproduce**
1. Open the editor
2. Double click on a material
3. Notice the issue

**Expected behavior**
The asset view should be focused when double clicking on a material or when clicking "Edit" in the material contextual menu.

**Screenshots**

https://github.com/user-attachments/assets/328d13c3-8913-408c-a495-5ee73c649962

Rishabh-Sukhwani:
Fix: Make Asset View focused when editing a material (#474 )

**Problem:**

When you double-clicked a material or right-clicked and chose "Edit", the Material Editor would open, but the focus would stay on it instead of switching back to the Asset View panel. This wasn't the expected behavior.

**Description:**

- Found the code in the AssetBrowser.cpp file that handles double-clicking a material and the "Edit" menu option.
- In both places, after the code that opens the Material Editor, added one extra line: EDITOR_PANEL(OvEditor::Panels::AssetView, "Asset View").Focus();
- This new line tells the editor to specifically put the focus back onto the Asset View panel right after the material is opened.
- Now, when you edit a material, the Asset View panel correctly gets focused.

**Screenshots:**

Asset-view-focus

**Related Issue:**

Closes #474

### DIFF
--- a/Sources/Overload/OvEditor/src/OvEditor/Panels/AssetBrowser.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Panels/AssetBrowser.cpp
@@ -798,7 +798,7 @@ namespace
 				if (material)
 				{
 					OpenInAssetView(*material);
-					OpenInMaterialEditor(*material);
+					EDITOR_EXEC(DelayAction([material]() { OpenInMaterialEditor(*material); }));
 				}
 			};
 
@@ -1264,7 +1264,7 @@ void OvEditor::Panels::AssetBrowser::ConsiderItem(OvUI::Widgets::Layout::TreeNod
 			{
 				auto& res = GetResource<OvCore::ResourceManagement::MaterialManager>(path, p_isEngineItem);
 				OpenInAssetView(res);
-				OpenInMaterialEditor(res);
+				EDITOR_EXEC(DelayAction([&res]() { OpenInMaterialEditor(res); }));
 			};
 		}
 


adriengivry (changes requested):
What's odd to me is that `OpenInAssetView(*material);` is already supposed to be setting the focus to the asset view:
```cpp
void OpenInAssetView(auto& p_resource)
{
	auto& assetView = EDITOR_PANEL(OvEditor::Panels::AssetView, "Asset View");
	assetView.SetResource(OvEditor::Panels::AssetView::ViewableResource{ &p_resource });
	assetView.Open();
	assetView.Focus(); // <-- here
}
```
But I think the issue is that we have 2 calls to `panel.Focus()`:
```cpp
OpenInAssetView(...); // <-- calls assetView.Focus();
OpenInMaterialEditor(...); // <-- calls materialEditor.Focus();
```
My guess is that only the last call to `Focus()` during a given frame will be executed. So with your proposed fix, the asset view will get focused, but the material editor won't anymore.

What I suggest is forcing `OpenInMaterialEditor` to execute in the next frame. To do so, you can use:
```cpp
// For the 'Edit' button case
OpenInAssetView(*material);
EDITOR_EXEC(DelayAction([material]() { OpenInMaterialEditor(*material); }));

// For the double click case
OpenInAssetView(res);
EDITOR_EXEC(DelayAction([&res]() { OpenInMaterialEditor(res); }));
```

Rishabh-Sukhwani:
Ah got it, I'll try this, thanks!

Rishabh-Sukhwani:
Asset-view-focus-new

I have changed the code as requested, it now seems to work fine as described. Both the Asset View and the Material Editor are opening now. I have amended my previous commit.

Thank you for the explanation! @adriengivry

adriengivry (approved):

PR #483, opened 2025-05-02
[WIP] Fixed grid blending

### LINKED ISSUE #380: Entities with alpha blending prevents the grid from rendering behind them
**Description**
Entities rendered with a material using alpha blending prevent the grid from rendering properly.

**To Reproduce**
1. Create a material using transparency
2. Look at the ambient smoke near the lava
3. Notice the issue

**Expected behavior**
The grid should still be visible through transparent entities

**Screenshots**

Image
_Cargo `game.ovscene` lava smoke particles_

Image
_A semi-transparent red-cube_

adriengivry:
## Description
* Fixed grid blending
  * Moved grid render pass to right after opaques, so that transparent objects can properly blend with it
  * Removed unused stencil usage
* Added stencil clear to `DebugActorRenderPass` for outline rendering (was necessary since the clear was removed from the `GridRendererPass`)

## Additional Notes
A side effect of moving the `GridRenderPass` to `Opaque + 1` instead of `Debug` (last), means that post-processing will apply to the grid, including FXAA and auto-exposure. In Cargo, this results in the grid looking brighter. However, since grid color can be tweaked from editor settings, this shouldn't be too much of an issue. It also helps the grid to better fit in within the scene. An issue with this approach is that colored lines might be affect by post-process effects, so they might not reflect the primary colors (red, greed, blue)

## To-Do
- [ ] Address issue in Fig 4. (axis lines colors affected by post-processing effects). We could separate the grid rendering into 2 passes: GridRenderPass (Opaque + 1), AxisIndicator (Debug)

## Related Issues
Fixes #380 

## Screenshots
image
_Fig 1. Cargo Game_

image
_Fig 2. Empty scene with post-processing and a white cube_

image
_Fig 3. Empty scene with post-processing and a white cube **after** adjusting scene view colors_

image
_Fig 4. When post-processing effects are enabled, the colored lines don't match the primary colors_

### DIFF
--- a/Sources/Overload/OvEditor/src/OvEditor/Rendering/DebugSceneRenderer.cpp
+++ b/Sources/Overload/OvEditor/src/OvEditor/Rendering/DebugSceneRenderer.cpp
@@ -173,6 +173,9 @@ protected:
 	{
 		TracyGpuZone("DebugActorRenderPass");
 
+		// Clear stencil buffer for outline rendering
+		m_renderer.Clear(false, false, true);
+
 		auto& debugSceneDescriptor = m_renderer.GetDescriptor<OvEditor::Rendering::DebugSceneRenderer::DebugSceneDescriptor>();
 
 		if (debugSceneDescriptor.selectedActor)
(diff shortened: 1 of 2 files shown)


adriengivry:
Addressed by https://github.com/Overload-Technologies/Overload/pull/511

PR #484, opened 2025-05-02
Materials can now be dropped onto an actor

### LINKED ISSUE #97: Material Drag & Drop onto Actor
Sometimes, it happens to be tedious to set a material to an `Actor` in the scene

**Expected solution**
It could be nice to be able to drag&drop a material from the asset browser to an `Actor` (in the `SceneView` or `Hierarchy`)

**Exemple**
drag_material

adriengivry:
## Description
Materials can now be dropped onto an actor in the scene view!

## Review Guidance
- Cleaned up the picking code so that it can be reused to detect what actor is under the mouse cursor outside of the picking handling for selection.
- Added proper methods to handle the different drop actions (`OnSceneDropped`, `OnModelDropped`, `OnMaterialDropped`)

## Limitations
- Only the material at index `0` in the `MaterialRenderer` will be affected.
- No preview of which object is going to be affected.
- No preview of the material on the object before it's applied (like Unity would do). The UX wouldn't be so great if we added that, since we only load resource synchronously. This would lead to the editor stuttering before dropping.

## Related Issues
Fixes #97 

## Screenshots

https://github.com/user-attachments/assets/73fcdde9-1f96-4050-ad84-00f17fe683c1

### DIFF
--- a/Sources/Overload/OvEditor/include/OvEditor/Panels/SceneView.h
+++ b/Sources/Overload/OvEditor/include/OvEditor/Panels/SceneView.h
@@ -6,8 +6,9 @@
 
 #pragma once
 
-#include "OvEditor/Panels/AViewControllable.h"
-#include "OvEditor/Core/GizmoBehaviour.h"
+#include <OvEditor/Core/GizmoBehaviour.h>
+#include <OvEditor/Panels/AViewControllable.h>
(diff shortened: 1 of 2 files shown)


Synyproxy (approved):
👍

PR #626, opened 2025-12-03
[WIP] Added automatic DPI scaling to editor

adriengivry:
## Description

This PR enables the following ImGui config flags:
- `ImGuiConfigFlags_DpiEnableScaleViewports`
- `ImGuiConfigFlags_DpiEnableScaleFonts`

Note: ImGui mentions:
```cpp
ImGuiConfigFlags_DpiEnableScaleViewports= 1 << 14,  // [BETA: Don't use] FIXME-DPI: Reposition and resize imgui windows when the DpiScale of a viewport changed (mostly useful for the main viewport hosting other window). Note that resizing the main window itself is up to your application.
ImGuiConfigFlags_DpiEnableScaleFonts    = 1 << 15,  // [BETA: Don't use] FIXME-DPI: Request bitmap-scaled fonts to match DpiScale. This is a very low-quality workaround. The correct way to handle DPI is _currently_ to replace the atlas and/or fonts in the Platform_OnChangedViewport callback, but this is all early work in progress.
```
So we might need to do more investigation whether this is the right approach.

**Edit: It seems like these config options do not exist anymore on ImGui's latest.**
**Edit 2: These options are now in `ImGuiIO`, as `ConfigDpiScaleFonts` and `ConfigDpiScaleViewports`**

**To-Do:**
- [ ] Fix some buttons don't scale properly
- [ ] Validate on a variety of screen sizes

## Related Issue(s)

Fixes #(issue number)

## Review Guidance

Write here.

## Screenshots/GIFs

**MacBook Pro M1 (Asahi Linux) 3456x2160**
| Before | After |
| - | - |
| image | image |

### DIFF
--- a/Sources/OvUI/include/OvUI/Core/UIManager.h
+++ b/Sources/OvUI/include/OvUI/Core/UIManager.h
@@ -96,6 +96,12 @@ namespace OvUI::Core
 		*/
 		void EnableDocking(bool p_value);
 
+		/**
+		* Enable DPI scaling
+		* @param p_value
+		*/
+		void EnableDPIScaling(bool p_value);
+
         /**
         * Reset the UI layout to the given configuration file
         * @param p_config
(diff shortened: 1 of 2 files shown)


adriengivry:
Seems like this should be handled fairly well with newer versions of imgui!
https://github.com/ocornut/imgui/blob/2cbdb7a337af7da5cfe6adf320f0e84ef56bee01/docs/FONTS.md?plain=1#L84

We should probably look into upgrading ImGUI to its latest version and use this feature.

Follow-up issue: https://github.com/Overload-Technologies/Overload/issues/713

PR #636, opened 2026-01-18
Update GLBackend - Fix stack smashing

### LINKED ISSUE #634: Stack smashing (buffer overflow) detected in debug builds
On Linux, when using the ./Build.sh shell script with the debug configuration, OvEditor opens but then crashes; this does not happen if the configuration is set to release version.

It's probably a problem with Make.

MunFaill:
## Description
Fixing the stack smashing that occurs when compiling with the debug configuration by setting a default integer on the glGetIntegerv params.

## Related Issue(s)
Fixes #634

## Review Guidance
GOTO: Sources/OvRendering/src/OvRendering/HAL/OpenGL/GLBackend.cpp.
Changed functions: int GetInt(uint32_t p_parameter) and int GetInt(uint32_t p_parameter, uint32_t p_index).
Lines: ~90 and  ~100

## Checklist

- [x] I have updated the documentation accordingly

### DIFF
--- a/Sources/OvRendering/src/OvRendering/HAL/OpenGL/GLBackend.cpp
+++ b/Sources/OvRendering/src/OvRendering/HAL/OpenGL/GLBackend.cpp
@@ -87,18 +87,21 @@ namespace
 		return static_cast<bool>(result);
 	}
 
-	int GetInt(uint32_t p_parameter)
+	int GetInt(uint32_t p_parameter) 
 	{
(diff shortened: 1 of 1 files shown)


adriengivry:
Thanks for submitting this PR! I only peaked at it so far, and I'll need to test it locally. Ideally we want a permanent fix rather than a workaround. I'll circle back with more feedback later. Thanks again!

PR #638, opened 2026-02-01
Fix buffer overflow in GLBackend

### LINKED ISSUE #634: Stack smashing (buffer overflow) detected in debug builds

MunFaill:
### Fix for Buffer Overflow by limiting the number of integers in GLint:
```cpp
//GLBackend.cpp

GLint result[4]; // Fixed size of 4

glGetIntegerv(p_parameter, result); // The max size of glGetIntegerv is 4

return static_cast<int>(result[0]);
```
## Related Issue(s)
Fixes #634 

## Checklist

## See also

[Why a fixed size of 4?](https://stackoverflow.com/a/19016823)


adriengivry (changes requested):
See comment above.

> [!NOTE]
> You can keep this PR open and push the edits to your branch. No need to create a new PR!

PR #641, opened 2026-02-01
Fix Dedicated GPU issue on linux

### LINKED ISSUE #639: USE_DEDICATED_GPU doesn't work on Linux.
## Description
`USE_DEDICATED_GPU` doesn't work very well on Linux, kernel 6.xx; my recommendation is to use `setenv` from the C standard library.

In my case, after the build, when running the 'nvidia-smi' command in the terminal, I could verify that OvEditor and OvGame were not using the dedicated graphics card.

MunFaill:
### Fix USE_DEDICATED_GPU on Linux:

```cpp
// Defines.h

// Linux version - use setenv function
#define FORCE_DEDICATED_GPU \
{\
    setenv("__NV_PRIME_RENDER_OFFLOAD", "1", 1);\
    setenv("__GLX_VENDOR_LIBRARY_NAME", "nvidia", 1);\
    setenv("DRI_PRIME", "1", 1);\
}
```
>[!note]
>"DRI_PRIME", "1", 1 is for dedicated GPUs only; if it's an integrated GPU, this line is ignored.

Changes in both Main.cpp (Game/Editor)

```cpp
// Outiside int main() [...]

#ifdef _WIN32
FORCE_DEDICATED_GPU
#endif

// Inside int main() [...]

#ifdef __linux__
FORCE_DEDICATED_GPU
#endif
```
## Related Issue(s)
Fixes #639

## Checklist

### DIFF
--- a/Sources/OvEditor/src/OvEditor/Main.cpp
+++ b/Sources/OvEditor/src/OvEditor/Main.cpp
@@ -20,12 +20,11 @@
 
 #ifdef _WIN32
 #undef APIENTRY
+FORCE_DEDICATED_GPU
 #include "Windows.h"
 #else
 #include <cstdlib>
 #endif
-
-FORCE_DEDICATED_GPU
 
 namespace
 {
@@ -80,7 +79,10 @@ namespace
 int main(int argc, char** argv)
 {
 	// UpdateWorkingDirectory(argv[0]);
-
+	#ifdef __linux__
+	FORCE_DEDICATED_GPU
+	#endif
+	
 	OvEditor::Settings::EditorSettings::Load();
 
 	std::optional<std::filesystem::path> projectPath;


adriengivry (changes requested):

adriengivry:
> From my research, I've realized that the previous solution is only problematic with Linux using Nvidia; AMD/Intel work perfectly. This runs into the issue of Nvidia's somewhat unstable support on Linux.
> 
> I believe the best solution is to continue with the current method, without changes.
> 
> For troubleshooting purposes, the problem is resolved by forcing the GPU to work using Nvidia Prime, e.g., #$ prime-run ./OvEditor

I think we can close this for now. Users can either use:
```
prime-run ./OvEditor
```
or
```
DRI_PRIME=1 ./OvEditor
```
or (if their discrete GPU has index 0):
```
DRI_PRIME=0 ./OvEditor
```

PR #648, opened 2026-03-31
Fix Inputs.GetMousePos to return FVector2 instead of FVector3

### LINKED ISSUE #447: Investigate Lua `GetMousePos` function returning `FVector3` instead of expected `FVector2`
## Description
In `LuaGlobalBindings.cpp`, it seems like `GetMousePos` is creating a `FVector3` instance out of 2 values (mouse position X and Y).

I'm wondering if that's necessary (workaround?), since `GetMousePos` is expected to return a `FVector2`.

## Expected Solution
- Test that lua scripts don't break when `GetMousePos` returns a `FVector2` instead of a `FVector3`
- Change `GetMousePos` to return a `FVector2`

Gopmyc:
## Summary

Inputs.GetMousePos currently returns an FVector3 while the scripting API documentation specifies a Vector2.

This pull request updates the Lua binding to return an FVector2, ensuring consistency with the documented behavior and with Inputs.GetMouseScroll.

## Changes

- Updated Inputs.GetMousePos in LuaGlobalBindings.cpp
- Replaced FVector3 with FVector2

## Testing

The project builds successfully after the change.

A Lua test script was used to observe the returned value:

```lua
return {
    OnStart = function()
        Debug.LogInfo(
            tostring(Inputs:GetMousePos())
        )
    end,
}
```

Before:
- Returned a Vector3 (x, y, z)

After:
- Returns a Vector2 (x, y)

## Related issue

Closes #447

### DIFF
--- a/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaGlobalBindings.cpp
+++ b/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaGlobalBindings.cpp
@@ -198,7 +198,7 @@ void BindLuaGlobal(sol::state& p_luaState)
 		"GetMouseButton", [](EMouseButton p_button) { return OVSERVICE(InputManager).GetMouseButtonState(p_button) == EMouseButtonState::MOUSE_DOWN; },
 		"GetMousePos", []() {
 			const auto mousePos = OVSERVICE(InputManager).GetMousePosition();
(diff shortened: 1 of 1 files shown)


adriengivry (approved):
Looks good to me! This could be a breaking change if any script tries to access the `z` component of the vector returned by `Inputs.GetMousePos()`, but since the `z` component wasn't intended to be there, I'm fine with the consequences.

Approved conditionally, subject to the build pipeline passing successfully

PR #667, opened 2026-04-08
Implemented model material/texture import and reuse on model/scene load

### LINKED ISSUE #303: Extract textures from fbx files
When importing an FBX model, it's preferable to have the option to extract texture files as well.

Gopmyc:
## Description
This PR implements automatic material/texture import and reuse for models, both when adding a model to the scene and when loading a `.ovscene`.

Main changes:
- Added model material import pipeline in editor actions (Assimp-based).
- Generated assets are stored in per-model folders:
  - `Assets/Materials/<ModelName>/`
  - `Assets/Textures/<ModelName>/`
- Reuse existing generated materials/textures instead of regenerating when already present.
- Applied the same generation/reuse logic when loading a scene from disk (`.ovscene`).
- Refreshed Asset Browser automatically when new assets are generated.
- Synced regenerated materials across all scene actors using the same model (avoids stale pink/error materials).
- Fixed picking pass skinning detection to use drawable skinning descriptors (compatible with `IMesh` usage).

## Related Issue(s)
Fixes #303

## Review Guidance
- Drag and drop a model (`.fbx`, etc.) into the scene:
  - Actor should be created with generated/assigned materials.
  - Materials/textures should be generated under model-specific subfolders.
- Re-import the same model:
  - Existing generated assets should be reused.
- Delete generated materials, then re-add/reload the model:
  - Materials should be regenerated and reassigned.
  - Other actors already using the same model should update as well.
- Load a `.ovscene` containing models:
  - Missing materials/textures should be generated/reused and assigned automatically.
- Verify picking still works correctly on skinned meshes after the picking-pass adjustment.

## Checklist
- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~

> The reason for the defensive handling is based on Assimp’s own texture semantic docs:
> 
> - Assimp explicitly states that mapping texture usages to `aiTextureType` is a best-effort process across DCC/tools/importers:
>   https://github.com/assimp/assimp/blob/master/include/assimp/material.h#L174-L181
> - `aiTextureType_HEIGHT` and `aiTextureType_NORMALS` are separate semantics in the API, but importers/exporters can still produce ambiguous mappings in practice:
>   https://github.com/assimp/assimp/blob/master/include/assimp/material.h#L232-L240
> 
> So in this PR I handle it defensively:
> - parse normal + height/displacement candidates,
> - if height resolves to the same texture as normal (or acts as fallback normal), do not enable parallax,
> - only enable parallax when a distinct height/displacement map is actually present._


adriengivry (changes requested):
Thanks for submitting this PR!

I can see a few major issues with this work:

1. **Assimp is being used directly in `OvEditor`.** Implementation details related to model loading should remain hidden from the application behind the `IModelParser` interface. Additional model data (i.e., material and texture information) should be fully handled by the `AssimpParser`.

2. **Project file pollution.** This PR assumes that the user wants to place automatically generated materials and textures in `Materials/{model_name}/` and `Models/{model_name}/`, respectively.

3. **Imported materials and models collision.** In some cases, if the user has multiple models with the same name in different folders, only the first imported model will properly generate materials and textures.

Ultimately, my vision for this feature is to offer something similar to what Unity does. I see models (FBX, OBJ, etc.) as “folders” that can be expanded to reveal all their embedded assets, including materials and potentially textures, if any.

Image

These embedded assets would be non-modifiable, as they would be generated when the model is loaded. They could then be dragged and dropped anywhere a `material` or `texture` is required. Paths for embedded assets could look like:

```
path/to/model.fbx:embedded_material.ovmat
path/to/model.fbx:embedded_texture.jpeg
```

In the model asset properties, we could add a toggle (above the model parsing flags) to determine whether embedded materials and textures should be generated.

This would also require updating the `PropagateFileRename` implementation to properly handle embedded assets.

Once this is fully implemented, I believe the “Generate materials” option in the model contextual menu could be safely removed.

The final step would be to add support for externally linked textures. Some models reference textures using relative paths, and it would be beneficial to support these as well. Generated materials should be able to reference linked textures correctly.

Given the scope of this work, I understand if it’s not manageable for you at the moment. We’ve postponed this effort for many years because it’s fairly complex and has a lot of dependencies.

Gopmyc:
Challenge accepted

Gopmyc:
> Thanks for submitting this PR!
> 
> I can see a few major issues with this work:
> 
> 1. **Assimp is being used directly in `OvEditor`.** Implementation details related to model loading should remain hidden from the application behind the `IModelParser` interface. Additional model data (i.e., material and texture information) should be fully handled by the `AssimpParser`.
> 2. **Project file pollution.** This PR assumes that the user wants to place automatically generated materials and textures in `Materials/{model_name}/` and `Models/{model_name}/`, respectively.
> 3. **Imported materials and models collision.** In some cases, if the user has multiple models with the same name in different folders, only the first imported model will properly generate materials and textures.
> 
> Ultimately, my vision for this feature is to offer something similar to what Unity does. I see models (FBX, OBJ, etc.) as “folders” that can be expanded to reveal all their embedded assets, including materials and potentially textures, if any.
> 
> Image
> These embedded assets would be non-modifiable, as they would be generated when the model is loaded. They could then be dragged and dropped anywhere a `material` or `texture` is required. Paths for embedded assets could look like:
> 
> ```
> path/to/model.fbx:embedded_material.ovmat
> path/to/model.fbx:embedded_texture.jpeg
> ```
> 
> In the model asset properties, we could add a toggle (above the model parsing flags) to determine whether embedded materials and textures should be generated.
> 
> This would also require updating the `PropagateFileRename` implementation to properly handle embedded assets.
> 
> Once this is fully implemented, I believe the “Generate materials” option in the model contextual menu could be safely removed.
> 
> The final step would be to add support for externally linked textures. Some models reference textures using relative paths, and it would be beneficial to support these as well. Generated materials should be able to reference linked textures correctly.
> 
> Given the scope of this work, I understand if it’s not manageable for you at the moment. We’ve postponed this effort for many years because it’s fairly complex and has a lot of dependencies.

I understand the direction and I agree with the overall idea, especially the Unity-like approach with embedded assets and keeping things clean through the parser.

That said, I don’t think it makes sense to include material generation and the embedded asset view in this PR. The scope gets pretty large and touches a lot of different parts of the engine (asset system, file handling, UI, parser responsibilities, etc.). To keep things clear and easier to review, I think it’s better to split this into multiple smaller PRs.

For now, I’ll focus on loading materials virtually through the parser and assigning them to the models at runtime. This should already improve rendering quality and material support, without polluting the project’s asset folders or introducing persistence/UI concerns too early.

Once that part is solid, we can move forward with separate PRs for embedded asset handling, optional generation, rename propagation, and support for external textures.

I’m going to reopen this PR and continue working on it with this scope.

adriengivry:
@Gopmyc 

Yep, this is a pretty massive change. Totally agree about breaking it down.
Could you explain more in detail what you mean by:
> For now, I’ll focus on loading materials virtually through the parser and assigning them to the models at runtime.

Gopmyc:
> @Gopmyc
> 
> Yep, this is a pretty massive change. Totally agree about breaking it down. Could you explain more in detail what you mean by:
> 
> > For now, I’ll focus on loading materials virtually through the parser and assigning them to the models at runtime.

Great question.

By “loading materials virtually through the parser”, I mean:
- no `.ovmat` / texture files are generated on disk in `Assets/`;
- `AssimpParser` extracts embedded material/texture data and stores it in the `Model` resource in memory;
- these are exposed through virtual paths like `model.fbx:embedded_material_0.ovmat` and `model.fbx:embedded_texture_0.png`.

Then, “assigning them at runtime” means `MaterialManager` / `TextureManager` resolve those virtual paths on demand, and renderer components assign them automatically when a model is loaded (or scene deserialized), especially for empty/default slots.

adriengivry (changes requested):
I didn't look too deep into the implementation that much yet, but I tested it locally, and here is my feedback:

**Issues/bugs:**
- `GENERATE_EMBEDDED_ASSETS` doesn't seem to work. Even when I set it to false the materials are still being loaded.
- Embedded textures don't seem to work (e.g. with this [ShaderBall.fbx](https://files.adrien-givry.com/api/public/dl/DpyKyfWx/public/Shader_Ball.fbx), loading it into Blender or Unity, you'll see a grid pattern on the ball, but you won't in Overload)
- Reloading linked (external) textures doesn't apply changes, but reloading the model after modifying a linked texture does work. This means that linked textures are actually duplicated in GPU memory (loaded twice). You can reproduce that issue with [Sponza FBX with linked textures](https://files.adrien-givry.com/share/qI6WwZ3g). If you import all the assets into a project, and load the FBX, and then modify a texture and reload it (right click on the texture > reload), you'll notice that the texture isn't updated on the model. To reload it you would have to reload the model instead.
- Parallax mapping seem to be enabled even when no height map is used. This can be reproduced by loading [Sponza FBX with linked textures](https://files.adrien-givry.com/share/qI6WwZ3g) and looking at textures from close, from an angle.
- We should also keep the "Generate Materials" for now, as it's still useful to change the appearance of a model. I know I mentioned we should remove it at some point, but I think it still serves a valid purpose for now.

**Notes:**
- Linked textures seem to work (e.g. [Sponza FBX with linked textures](https://files.adrien-givry.com/share/qI6WwZ3g), loading the FBX will properly find the textures using the relative path. Since you mentioned this, I thought this was gonna be in another PR, but I'm ok with external textures being supported in this PR as well:
  > Once that part is solid, we can move forward with separate PRs for embedded asset handling, optional generation, rename propagation, and support for external textures.

Overall this seem feature complete (once it will be fully fixed). I like the scope of the PR, which I would summarize as:
- Added support to load materials embedded into a model
- Added support to load textures embedded into a model and automatically attaching them to the correct embedded materials
- Added support to attach externally linked textures to embedded materials

Gopmyc:
> I didn't look too deep into the implementation that much yet, but I tested it locally, and here is my feedback:
> 
> **Issues/bugs:**
> 
> * `GENERATE_EMBEDDED_ASSETS` doesn't seem to work. Even when I set it to false the materials are still being loaded.
> * Embedded textures don't seem to work (e.g. with this [ShaderBall.fbx](https://files.adrien-givry.com/api/public/dl/DpyKyfWx/public/Shader_Ball.fbx), loading it into Blender or Unity, you'll see a grid pattern on the ball, but you won't in Overload)
> * Reloading linked (external) textures doesn't apply changes, but reloading the model after modifying a linked texture does work. This means that linked textures are actually duplicated in GPU memory (loaded twice). You can reproduce that issue with [Sponza FBX with linked textures](https://files.adrien-givry.com/share/qI6WwZ3g). If you import all the assets into a project, and load the FBX, and then modify a texture and reload it (right click on the texture > reload), you'll notice that the texture isn't updated on the model. To reload it you would have to reload the model instead.
> * Parallax mapping seem to be enabled even when no height map is used. This can be reproduced by loading [Sponza FBX with linked textures](https://files.adrien-givry.com/share/qI6WwZ3g) and looking at textures from close, from an angle.
> * We should also keep the "Generate Materials" for now, as it's still useful to change the appearance of a model. I know I mentioned we should remove it at some point, but I think it still serves a valid purpose for now.
> 
> **Notes:**
> 
> * Linked textures seem to work (e.g. [Sponza FBX with linked textures](https://files.adrien-givry.com/share/qI6WwZ3g), loading the FBX will properly find the textures using the relative path. Since you mentioned this, I thought this was gonna be in another PR, but I'm ok with external textures being supported in this PR as well:
>   > Once that part is solid, we can move forward with separate PRs for embedded asset handling, optional generation, rename propagation, and support for external textures.
> 
> Overall this seem feature complete (once it will be fully fixed). I like the scope of the PR, which I would summarize as:
> 
> * Added support to load materials embedded into a model
> * Added support to load textures embedded into a model and automatically attaching them to the correct embedded materials
> * Added support to attach externally linked textures to embedded materials

Thanks for the detailed review, this was very helpful.

I addressed each issue you listed:

- `GENERATE_EMBEDDED_ASSETS` now behaves correctly: when disabled, embedded materials are no longer applied (fallback is used instead of stale embedded refs).
- Embedded textures are now properly resolved, including named embedded texture references (not only `*0`, `*1`, etc.).
- Linked external textures are now bound through their real shared texture resource path, so reloading a linked texture updates model materials correctly and avoids duplicate GPU loads.
- `PARALLAX_MAPPING` is now enabled only when a valid height map is actually bound.
- I kept the **"Generate materials..."** model context-menu entry.

I also re-validated with ShaderBall/Sponza-style cases, and the PR scope is now aligned with your summary:
- embedded materials loading
- embedded textures loading and auto-assignment
- externally linked textures attached to embedded materials

adriengivry (changes requested):
Thanks for updating the PR! It looks really good, and seem to work in most cases!

One issue I found is with externally linked textures (e.g. [Sponza FBX with linked textures](https://files.adrien-givry.com/share/qI6WwZ3g)) external textures are now properly being used (yay! 🎉) but parallax mapping still look like it's enabled even though no height map is provided.

Image

Gopmyc:
> Thanks for updating the PR! It looks really good, and seem to work in most cases!
> 
> One issue I found is with externally linked textures (e.g. [Sponza FBX with linked textures](https://files.adrien-givry.com/share/qI6WwZ3g)) external textures are now properly being used (yay! 🎉) but parallax mapping still look like it's enabled even though no height map is provided.
> 
> Image

Thanks for the feedback !

I just pushed a fix for this point, and it now seems corrected on my side :
parallax is no longer enabled for externally linked textures when no real height map is provided (Sponza case).

I adjusted the material extraction/runtime binding logic to avoid treating height/bump data as parallax input in this scenario.

Sources/OvCore/src/OvCore/ECS/Components/CMaterialRenderer.cpp
```diff
@@ -13,11 +15,17 @@
...
 #include <OvCore/ResourceManagement/MaterialManager.h>
+#include <OvRendering/Resources/Parsers/EmbeddedAssetPath.h>
 
 #include <OvUI/Widgets/Layout/Dummy.h>
 #include <OvUI/Widgets/Texts/Text.h>
 #include <OvUI/Widgets/Visual/Separator.h>
 
+namespace
+{
+	constexpr std::string_view kDefaultMaterialPath = ":Materials\\Default.ovmat";
```
adriengivry: Ideally we don't want to reference assets by path outside of `OvEditor`. It should be the responsability of the editor to decide which material should be considered the default material. In this case, we could be using `nullptr` (no material) if no embedded material exists. The editor could then fill in the blank when needed.

Sources/OvCore/src/OvCore/ResourceManagement/MaterialManager.cpp
```diff
@@ -6,8 +6,206 @@
...
+#include <OvCore/ResourceManagement/ModelManager.h>
+#include <OvCore/ResourceManagement/ShaderManager.h>
+#include <OvCore/ResourceManagement/TextureManager.h>
+#include <OvRendering/Resources/Parsers/EmbeddedAssetPath.h>
+#include <OvTools/Utils/PathParser.h>
+
+namespace
+{
+	constexpr std::string_view kDefaultMaterialPath = ":Materials\\Default.ovmat";
+	constexpr std::string_view kStandardShaderPath = ":Shaders\\Standard.ovfx";
```
adriengivry: Same as above, `OvCore` shouldn't reference assets directly. Maybe the `MaterialManager` could be updated so that the editor and game could provide a reference to the "Standard" shader as well as its expected uniform names to use when creating embedded materials.
For instance:
```cpp
struct ShaderDefinition
{
    std::string shaderPath;
    std::string albedoPropertyName;
    std::string metallicPropertyName;
    // ...
}
```
And the editor and game after creating the `MaterialManager` could:
```cpp
materialManager->ProvideStandardShaderDefinition({
    .shaderPath = ":Shaders/Standard.ovfx",
    .albedoPropertyName = "u_Albedo,
    // ...
});
```

adriengivry: I'm open to other architectures/designs, but we really should keep the file paths outside of `OvCore`.

Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
```diff
@@ -49,10 +53,60 @@ OvRendering::Resources::Parsers::EModelParserFlags GetAssetMetadata(const std::s
 	return flags;
 }
 
+bool ShouldGenerateEmbeddedAssets(const std::string& p_path)
```
adriengivry: Should be in an anonymous namespace to avoid polluting the global namespace.

Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
```diff
@@ -49,10 +53,60 @@ OvRendering::Resources::Parsers::EModelParserFlags GetAssetMetadata(const std::s
...
+		if (!p_parseIndex(embeddedAssetPath->assetName))
+		{
+			continue;
+		}
+
+		p_resourceManager.AResourceManager::ReloadResource(resourcePath);
+	}
+}
+
+void ReloadEmbeddedModelResources(const std::string& p_modelPath)
```
adriengivry: Same as above, should be in an anonymous namespace.

Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
```diff
@@ -6,6 +6,10 @@
 
 #include "OvCore/ResourceManagement/ModelManager.h"
 
+#include <OvCore/Global/ServiceLocator.h>
+#include <OvCore/ResourceManagement/MaterialManager.h>
+#include <OvCore/ResourceManagement/TextureManager.h>
+#include <OvRendering/Resources/Parsers/EmbeddedAssetPath.h>
 #include <OvTools/Filesystem/IniFile.h>
 
 OvRendering::Resources::Parsers::EModelParserFlags GetAssetMetadata(const std::string& p_path)
```
adriengivry: This should be updated to return a
```cpp
struct ModelMetaData
{
    OvRendering::Resources::Parsers::EModelParserFlags parserFlags;
    bool generateEmbeddedAssets;
}
```
This way `GetAssetMetadata` would return all the metadata needed for the loader to process.

Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
adriengivry: Actually could be removed completely  once `struct ModelMetaData` is added, and `GetAssetMetadata` updated.

Sources/OvCore/src/OvCore/ResourceManagement/TextureManager.cpp
```diff
@@ -22,6 +28,21 @@ namespace
...
+		using enum ETextureFilteringMode;
+		using enum ETextureWrapMode;
+
+		return TextureMetaData{
+			.minFilter = LINEAR_MIPMAP_LINEAR,
+			.magFilter = LINEAR,
+			.horizontalWrap = REPEAT,
+			.verticalWrap = REPEAT,
+			.generateMipmap = true
+		};
```
adriengivry: This is redundant with `LoadTextureMetadata`, default values are being copied. We should make sure default values aren't defined multiple times.

Sources/OvCore/src/OvCore/ResourceManagement/TextureManager.cpp
```diff
@@ -38,10 +59,206 @@ namespace
...
+
+		OvRendering::Resources::Texture* texture = nullptr;
+
+		switch (textureData.sourceType)
+		{
+		case SourceType::EXTERNAL_FILE:
+			if (!textureData.sourcePath.empty())
+			{
+				const auto realTexturePath = ResolveRuntimeExternalTexturePath(textureData.sourcePath, p_modelRealPath);
+				texture = OvRendering::Resources::Loaders::TextureLoader::Create(
```
adriengivry: What happens if the texture at this path is already loaded and exposed in the texture manager? Will it be loaded twice?

adriengivry: My guess is that this code path (`case SourceType::EXTERNAL_FILE`) is never called, and should probably never get called.

When texture data comes from an external file, there should be no need for additional resource creation.

Maybe even putting an `OVASSERT` in this code path would make sense.

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -89,9 +263,16 @@ void OvEditor::Core::EditorActions::SaveSceneToDisk(OvCore::SceneSystem::Scene&
...
 {
 	if (GetCurrentEditorMode() != EEditorMode::EDIT)
+	{
 		StopPlaying();
+	}
 
-	m_context.sceneManager.LoadScene(p_path, p_absolute);
+	if (!m_context.sceneManager.LoadScene(p_path, p_absolute))
+	{
+		return;
```
adriengivry, line 176: I appreciate the addition! Since we check if the scene loaded properly, shouldn't we display an error message in case it failed?

Sources/OvRendering/src/OvRendering/Resources/Loaders/TextureLoader.cpp
```diff
@@ -71,7 +103,7 @@ namespace
 			.useMipMaps = p_generateMipmap
 		});
 
-		p_texture.Upload(p_data, EFormat::RGBA, p_hdr ? EPixelDataType::FLOAT : EPixelDataType::UNSIGNED_BYTE);
+		p_texture.Upload(const_cast<void*>(p_data), EFormat::RGBA, p_hdr ? EPixelDataType::FLOAT : EPixelDataType::UNSIGNED_BYTE);
```
adriengivry: Why was `p_data` made `const` in the first place? `const_cast` is dangerous and should be avoided in this context (fine to keep with resource paths).

Sources/OvRendering/include/OvRendering/Resources/Model.h
```diff
@@ -51,6 +52,26 @@ namespace OvRendering::Resources
...
+
+		/**
+		* Returns embedded texture data extracted from the model
+		*/
+		const std::vector<EmbeddedTextureData>& GetEmbeddedTextures() const;
+
+		/**
+		* Returns an embedded material data by index
+		*/
+		const EmbeddedMaterialData* GetEmbeddedMaterial(uint32_t p_index) const;
```
adriengivry: Prefer `OvTools::Utils::OptRef<const EmbeddedMaterialData>`  instead of `const EmbeddedMaterialData*`

Sources/OvRendering/include/OvRendering/Resources/Model.h
```diff
@@ -51,6 +52,26 @@ namespace OvRendering::Resources
...
+
+		/**
+		* Returns an embedded texture data by index
+		*/
+		const EmbeddedTextureData* GetEmbeddedTexture(uint32_t p_index) const;
```
adriengivry: Same as above, prefer `OvTools::Utils::OptRef<const EmbeddedTextureData>`.

adriengivry (changes requested):
Only a few comment, nothing too structural 😄

Sources/OvCore/src/OvCore/ECS/Components/CMaterialRenderer.cpp
Gopmyc: Good call. I removed the default material path fallback from OvCore here, if no embedded material exists, the slot stays empty and Editor/Game decide what to use

Sources/OvCore/src/OvCore/ResourceManagement/MaterialManager.cpp
Gopmyc: Implemented this: MaterialManager now supports ProvideStandardShaderDefinition(...), and Editor/Game provide the Standard shader + uniform/feature names

Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
Gopmyc: Done, moved to an anonymous namespace

Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
Gopmyc: Done as well, now in anonymous namespace too

Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
Gopmyc: Implemented: GetAssetMetadata now returns a ModelMetaData struct (parserFlags + generateEmbeddedAssets)

Sources/OvCore/src/OvCore/ResourceManagement/TextureManager.cpp
Gopmyc: You’re right. I removed that external-file embedded creation path to avoid duplicate texture loads

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: Done. On scene load failure we now log an error and show an editor message box

Sources/OvRendering/src/OvRendering/Resources/Loaders/TextureLoader.cpp
Gopmyc: Agreed. I removed that const_cast from the texture upload path

Sources/OvRendering/include/OvRendering/Resources/Model.h
Gopmyc: Updated to OptRef<const EmbeddedMaterialData> as suggested

Sources/OvRendering/include/OvRendering/Resources/Model.h
Gopmyc: Updated this one too: OptRef<const EmbeddedTextureData>

Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
```diff
@@ -12,100 +12,105 @@
...
-
-	return flags;
-}
-
-bool ShouldGenerateEmbeddedAssets(const std::string& p_path)
-{
-	auto metaFile = OvTools::Filesystem::IniFile(p_path + ".meta");
-	return metaFile.GetOrDefault("GENERATE_EMBEDDED_ASSETS", true);
-}
+	struct ModelMetaData
```
adriengivry: nip: `ModelMetadata` instead of `ModelMetaData` to be consistent with the rest of the codebase

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -270,6 +270,16 @@ void OvEditor::Core::EditorActions::LoadSceneFromDisk(const std::string& p_path,
...
 	if (!m_context.sceneManager.LoadScene(p_path, p_absolute))
 	{
+		OVLOG_ERROR("Failed to load scene from disk: " + p_path);
+
+		OvWindowing::Dialogs::MessageBox message(
+			"Failed to Load Scene",
+			std::format("Could not load scene:\n{}", p_path),
+			OvWindowing::Dialogs::MessageBox::EMessageType::ERROR,
+			OvWindowing::Dialogs::MessageBox::EButtonLayout::OK,
+			true
```
adriengivry: `OVLOG_ERROR` is sufficient here, it should silently fail instead of showing an error message

Sources/OvCore/include/OvCore/ResourceManagement/MaterialManager.h
```diff
@@ -17,6 +20,37 @@ namespace OvCore::ResourceManagement
...
+		/**
+		* Provide the shader definition used to create embedded materials
+		* @param p_definition
+		*/
+		void ProvideStandardShaderDefinition(StandardShaderDefinition p_definition);
+
+		/**
+		* Returns the current shader definition used for embedded materials
+		*/
+		const StandardShaderDefinition* GetStandardShaderDefinition() const;
```
adriengivry: We should avoid pointers where we're not doing any manual memory management, `const std::optional<StandardShaderDefinition>&`  would work here.

adriengivry (changes requested):
Looking really good! Only have minor comments, I'll do a final full review pass soon

Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
Gopmyc: Done, I renamed it to ModelMetadata for consistency

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: Good point, I removed the popup and kept OVLOG_ERROR only

Sources/OvCore/include/OvCore/ResourceManagement/MaterialManager.h
Gopmyc: greed. I changed it to return const std::optional<StandardShaderDefinition>& and updated the call sites accordingly

Sources/OvCore/src/OvCore/ResourceManagement/TextureManager.cpp
```diff
@@ -22,26 +28,188 @@ namespace
 		bool generateMipmap;
 	};
 
-	TextureMetaData LoadTextureMetadata(const std::string_view p_filePath)
+	TextureMetaData GetDefaultTextureMetadata()
```
adriengivry: Not needed, as the default values can be built into the struct definition.
While you're at it, let's also rename `TextureMetaData` to `TextureMetadata` 😄

Sources/OvCore/src/OvCore/ResourceManagement/TextureManager.cpp
```diff
@@ -22,26 +28,188 @@ namespace
...
+			.magFilter = static_cast<OvRendering::Settings::ETextureFilteringMode>(metaFile.GetOrDefault("MAG_FILTER", static_cast<int>(defaultMetadata.magFilter))),
+			.horizontalWrap = static_cast<OvRendering::Settings::ETextureWrapMode>(metaFile.GetOrDefault("HORIZONTAL_WRAP", static_cast<int>(defaultMetadata.horizontalWrap))),
+			.verticalWrap = static_cast<OvRendering::Settings::ETextureWrapMode>(metaFile.GetOrDefault("VERTICAL_WRAP", static_cast<int>(defaultMetadata.verticalWrap))),
+			.generateMipmap = metaFile.GetOrDefault("ENABLE_MIPMAPPING", defaultMetadata.generateMipmap)
+		};
+	}
+
+	struct EmbeddedTextureContext
+	{
+		const OvRendering::Resources::EmbeddedTextureData* textureData = nullptr;
```
adriengivry: A valid `EmbeddedTextureContext` should never have a `textureData` with a `nullptr`  value.
We could change `textureData`'s type to `const OvRendering::Resources::EmbeddedTextureData&`

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -90,9 +264,17 @@ void OvEditor::Core::EditorActions::SaveSceneToDisk(OvCore::SceneSystem::Scene&
...
+	}
 
-	m_context.sceneManager.LoadScene(p_path, p_absolute);
+	if (!m_context.sceneManager.LoadScene(p_path, p_absolute))
+	{
+		OVLOG_ERROR("Failed to load scene from disk: " + p_path);
+		return;
+	}
+
+	ApplyEmbeddedModelMaterialsToCurrentScene(m_context);
```
adriengivry: I can't think of any case where this would be useful. Scenes should be loaded as is. If embedded materials are not in the scene data, they shouldn't be used.

This invocation as well as the entire `ApplyEmbeddedModelMaterialsToCurrentScene` function could be removed.

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -618,12 +800,27 @@ OvCore::ECS::Actor & OvEditor::Core::EditorActions::CreateActorWithModel(const s
...
-		materialRenderer.FillWithMaterial(*material);
+	const auto defaultMaterial = m_context.materialManager[std::string{ kDefaultMaterialPath }];
+	if (defaultMaterial)
+	{
+		materialRenderer.FillWithMaterial(*defaultMaterial);
+	}
+
+	if (model)
+	{
+		ApplyEmbeddedModelMaterials(
```
adriengivry: Instead of having this function in `EditorActions.cpp`, this should be added to `CMaterialRenderer` and named something like `FillWithEmbeddedMaterials()`. Maybe `CMaterialRenderer::ApplyEmbeddedMaterialFallback()` could even be removed? No need to pass the context, as it's accessible from `CMaterialRenderer` with the service locator. We could even just call `FillWithEmbeddedMaterials()` instead of `FillWithMaterial()` above, and let the function decide whether to use embedded materials or the default one based on availability.

Essentially the material renderer should have 2 methods:
- `FillWithMaterial`: fill all material entries on a material renderer with a given material (e.g. default)
- `FillWithEmbeddedMaterials`: fill all material entries on a material renderer with embedded materials coming from a given model, or the default material if no embedded ones are found. (a boolean argument could be added to choose whether we want to only add embedded materials to empty slots, which would correspond to what `ApplyEmbeddedMaterialFallback` was trying to do)

adriengivry (changes requested):

Sources/OvCore/src/OvCore/ResourceManagement/TextureManager.cpp
Gopmyc: Refactored: defaults are now in TextureMetadata, and TextureMetaData was renamed

Sources/OvCore/src/OvCore/ResourceManagement/TextureManager.cpp
Gopmyc: Agreed, textureData is now a reference instead of a nullable pointer

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: Removed this scene-load pass so scenes are now loaded exactly as serialized

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: Moved embedded material filling into CMaterialRenderer (FillWithEmbeddedMaterials) and removed the EditorActions helper

Sources/OvCore/src/OvCore/ECS/Components/CModelRenderer.cpp
```diff
@@ -23,10 +23,21 @@
 
 OvCore::ECS::Components::CModelRenderer::CModelRenderer(ECS::Actor& p_owner) : AComponent(p_owner)
 {
-	m_modelChangedEvent += [this]
+	m_modelChangedNotifier += [this]
+	{
+		m_modelChangedEvent.Invoke(true);
```
adriengivry:
```suggestion
		m_modelChangedEvent.Invoke(false);
```
This way, we don't override any material when changing the model in the "Model Renderer".

adriengivry (changes requested):
After reviewing the entire PR again, I don't see any use case for `SetModel(model, true)`. There is no code-path where we specifically call `SetModel()` with `p_updateMaterials` set to `true`, so we could remove it completely. This way we can get rid of the duplicated event (`m_modelChangedNotifier`), and simplify deserialization.

I've tried it locally and it works perfectly, here is the diff:
```
diff --git a/Sources/OvCore/include/OvCore/ECS/Components/CModelRenderer.h b/Sources/OvCore/include/OvCore/ECS/Components/CModelRenderer.h
index 7faa794..147da9b 100644
--- a/Sources/OvCore/include/OvCore/ECS/Components/CModelRenderer.h
+++ b/Sources/OvCore/include/OvCore/ECS/Components/CModelRenderer.h
@@ -51,9 +51,8 @@ namespace OvCore::ECS::Components
                /**
                * Defines the model to use
                * @param p_model
-               * @param p_updateMaterials
                */
-               void SetModel(OvRendering::Resources::Model* p_model, bool p_updateMaterials = false);
+               void SetModel(OvRendering::Resources::Model* p_model);

                /**
                * Returns the current model
@@ -104,7 +103,7 @@ namespace OvCore::ECS::Components

        private:
                OvRendering::Resources::Model* m_model = nullptr;
-               OvTools::Eventing::Event<bool> m_modelChangedEvent;
+               OvTools::Eventing::Event<> m_modelChangedEvent;
                OvTools::Eventing::Event<> m_modelChangedNotifier;
                OvRendering::Geometry::BoundingSphere m_customBoundingSphere = { {}, 1.0f };
                EFrustumBehaviour m_frustumBehaviour = EFrustumBehaviour::MESH_BOUNDS;
diff --git a/Sources/OvCore/src/OvCore/ECS/Components/CModelRenderer.cpp b/Sources/OvCore/src/OvCore/ECS/Components/CModelRenderer.cpp
index e1b0cb4..da5fbaa 100644
--- a/Sources/OvCore/src/OvCore/ECS/Components/CModelRenderer.cpp
+++ b/Sources/OvCore/src/OvCore/ECS/Components/CModelRenderer.cpp
@@ -23,19 +23,10 @@

 OvCore::ECS::Components::CModelRenderer::CModelRenderer(ECS::Actor& p_owner) : AComponent(p_owner)
 {
-       m_modelChangedNotifier += [this]
-       {
-               m_modelChangedEvent.Invoke(true);
-       };
-
-       m_modelChangedEvent += [this](bool p_updateMaterials)
+       m_modelChangedEvent += [this]()
        {
                if (auto materialRenderer = owner.GetComponent<CMaterialRenderer>())
                {
-                       if (p_updateMaterials)
-                       {
-                               materialRenderer->FillWithEmbeddedMaterials(true);
-                       }
                        materialRenderer->UpdateMaterialList();
                }

@@ -54,10 +45,10 @@ std::string OvCore::ECS::Components::CModelRenderer::GetTypeName()
        return std::string{ComponentTraits<CModelRenderer>::Name};
 }

-void OvCore::ECS::Components::CModelRenderer::SetModel(OvRendering::Resources::Model* p_model, bool p_updateMaterials)
+void OvCore::ECS::Components::CModelRenderer::SetModel(OvRendering::Resources::Model* p_model)
 {
        m_model = p_model;
-       m_modelChangedEvent.Invoke(p_updateMaterials);
+       m_modelChangedEvent.Invoke();
 }

 OvRendering::Resources::Model * OvCore::ECS::Components::CModelRenderer::GetModel() const
@@ -97,7 +88,7 @@ void OvCore::ECS::Components::CModelRenderer::OnDeserialize(tinyxml2::XMLDocumen
 {
        OvRendering::Resources::Model* model = nullptr;
        OvCore::Helpers::Serializer::DeserializeModel(p_doc, p_node, "model", model);
-       SetModel(model, false);
+       SetModel(model);

        OvCore::Helpers::Serializer::DeserializeInt(p_doc, p_node, "frustum_behaviour", reinterpret_cast<int&>(m_frustumBehaviour));
        OvCore::Helpers::Serializer::DeserializeVec3(p_doc, p_node, "custom_bounding_sphere_position", m_customBoundingSphere.position);
@@ -108,7 +99,7 @@ void OvCore::ECS::Components::CModelRenderer::OnInspector(OvUI::Internal::Widget
 {
        using namespace OvCore::Helpers;

-       GUIDrawer::DrawMesh(p_root, "Model", m_model, &m_modelChangedNotifier);
+       GUIDrawer::DrawMesh(p_root, "Model", m_model, &m_modelChangedEvent);

        GUIDrawer::CreateTitle(p_root, "Frustum Culling Behaviour");
        auto& boundingMode = p_root.CreateWidget<OvUI::Widgets::Selection::ComboBox>(0);
diff --git a/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaComponentsBindings.cpp b/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaComponentsBindings.cpp
index 8cd796d..9fbe84f 100644
--- a/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaComponentsBindings.cpp
+++ b/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaComponentsBindings.cpp
@@ -76,10 +76,7 @@ void BindLuaComponents(sol::state& p_luaState)
        p_luaState.new_usertype<CModelRenderer>("ModelRenderer",
                sol::base_classes, sol::bases<AComponent>(),
                "GetModel", &CModelRenderer::GetModel,
-               "SetModel", [](CModelRenderer& p_this, OvRendering::Resources::Model* p_model)
-               {
-                       p_this.SetModel(p_model);
-               },
+               "SetModel", &CModelRenderer::SetModel,
                "GetFrustumBehaviour", &CModelRenderer::GetFrustumBehaviour,
                "SetFrustumBehaviour", &CModelRenderer::SetFrustumBehaviour
        );
```

Gopmyc:
> After reviewing the entire PR again, I don't see any use case for `SetModel(model, true)`. There is no code-path where we specifically call `SetModel()` with `p_updateMaterials` set to `true`, so we could remove it completely. This way we can get rid of the duplicated event (`m_modelChangedNotifier`), and simplify deserialization.
> 
> I've tried it locally and it works perfectly, here is the diff:
> 
> ```
> diff --git a/Sources/OvCore/include/OvCore/ECS/Components/CModelRenderer.h b/Sources/OvCore/include/OvCore/ECS/Components/CModelRenderer.h
> index 7faa794..147da9b 100644
> --- a/Sources/OvCore/include/OvCore/ECS/Components/CModelRenderer.h
> +++ b/Sources/OvCore/include/OvCore/ECS/Components/CModelRenderer.h
> @@ -51,9 +51,8 @@ namespace OvCore::ECS::Components
>                 /**
>                 * Defines the model to use
>                 * @param p_model
> -               * @param p_updateMaterials
>                 */
> -               void SetModel(OvRendering::Resources::Model* p_model, bool p_updateMaterials = false);
> +               void SetModel(OvRendering::Resources::Model* p_model);
> 
>                 /**
>                 * Returns the current model
> @@ -104,7 +103,7 @@ namespace OvCore::ECS::Components
> 
>         private:
>                 OvRendering::Resources::Model* m_model = nullptr;
> -               OvTools::Eventing::Event<bool> m_modelChangedEvent;
> +               OvTools::Eventing::Event<> m_modelChangedEvent;
>                 OvTools::Eventing::Event<> m_modelChangedNotifier;
>                 OvRendering::Geometry::BoundingSphere m_customBoundingSphere = { {}, 1.0f };
>                 EFrustumBehaviour m_frustumBehaviour = EFrustumBehaviour::MESH_BOUNDS;
> diff --git a/Sources/OvCore/src/OvCore/ECS/Components/CModelRenderer.cpp b/Sources/OvCore/src/OvCore/ECS/Components/CModelRenderer.cpp
> index e1b0cb4..da5fbaa 100644
> --- a/Sources/OvCore/src/OvCore/ECS/Components/CModelRenderer.cpp
> +++ b/Sources/OvCore/src/OvCore/ECS/Components/CModelRenderer.cpp
> @@ -23,19 +23,10 @@
> 
>  OvCore::ECS::Components::CModelRenderer::CModelRenderer(ECS::Actor& p_owner) : AComponent(p_owner)
>  {
> -       m_modelChangedNotifier += [this]
> -       {
> -               m_modelChangedEvent.Invoke(true);
> -       };
> -
> -       m_modelChangedEvent += [this](bool p_updateMaterials)
> +       m_modelChangedEvent += [this]()
>         {
>                 if (auto materialRenderer = owner.GetComponent<CMaterialRenderer>())
>                 {
> -                       if (p_updateMaterials)
> -                       {
> -                               materialRenderer->FillWithEmbeddedMaterials(true);
> -                       }
>                         materialRenderer->UpdateMaterialList();
>                 }
> 
> @@ -54,10 +45,10 @@ std::string OvCore::ECS::Components::CModelRenderer::GetTypeName()
>         return std::string{ComponentTraits<CModelRenderer>::Name};
>  }
> 
> -void OvCore::ECS::Components::CModelRenderer::SetModel(OvRendering::Resources::Model* p_model, bool p_updateMaterials)
> +void OvCore::ECS::Components::CModelRenderer::SetModel(OvRendering::Resources::Model* p_model)
>  {
>         m_model = p_model;
> -       m_modelChangedEvent.Invoke(p_updateMaterials);
> +       m_modelChangedEvent.Invoke();
>  }
> 
>  OvRendering::Resources::Model * OvCore::ECS::Components::CModelRenderer::GetModel() const
> @@ -97,7 +88,7 @@ void OvCore::ECS::Components::CModelRenderer::OnDeserialize(tinyxml2::XMLDocumen
>  {
>         OvRendering::Resources::Model* model = nullptr;
>         OvCore::Helpers::Serializer::DeserializeModel(p_doc, p_node, "model", model);
> -       SetModel(model, false);
> +       SetModel(model);
> 
>         OvCore::Helpers::Serializer::DeserializeInt(p_doc, p_node, "frustum_behaviour", reinterpret_cast<int&>(m_frustumBehaviour));
>         OvCore::Helpers::Serializer::DeserializeVec3(p_doc, p_node, "custom_bounding_sphere_position", m_customBoundingSphere.position);
> @@ -108,7 +99,7 @@ void OvCore::ECS::Components::CModelRenderer::OnInspector(OvUI::Internal::Widget
>  {
>         using namespace OvCore::Helpers;
> 
> -       GUIDrawer::DrawMesh(p_root, "Model", m_model, &m_modelChangedNotifier);
> +       GUIDrawer::DrawMesh(p_root, "Model", m_model, &m_modelChangedEvent);
> 
>         GUIDrawer::CreateTitle(p_root, "Frustum Culling Behaviour");
>         auto& boundingMode = p_root.CreateWidget<OvUI::Widgets::Selection::ComboBox>(0);
> diff --git a/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaComponentsBindings.cpp b/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaComponentsBindings.cpp
> index 8cd796d..9fbe84f 100644
> --- a/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaComponentsBindings.cpp
> +++ b/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaComponentsBindings.cpp
> @@ -76,10 +76,7 @@ void BindLuaComponents(sol::state& p_luaState)
>         p_luaState.new_usertype<CModelRenderer>("ModelRenderer",
>                 sol::base_classes, sol::bases<AComponent>(),
>                 "GetModel", &CModelRenderer::GetModel,
> -               "SetModel", [](CModelRenderer& p_this, OvRendering::Resources::Model* p_model)
> -               {
> -                       p_this.SetModel(p_model);
> -               },
> +               "SetModel", &CModelRenderer::SetModel,
>                 "GetFrustumBehaviour", &CModelRenderer::GetFrustumBehaviour,
>                 "SetFrustumBehaviour", &CModelRenderer::SetFrustumBehaviour
>         );
> ```

Done, I removed SetModel(model, bool) entirely and simplified CModelRenderer as suggested.
SetModel is back to a single-arg API, the duplicated notifier/event flow is gone, and deserialization now just calls SetModel(model)

Sources/OvCore/src/OvCore/ECS/Components/CModelRenderer.cpp
Gopmyc: Applied.
Model changes from the inspector no longer trigger embedded material overwrite anymore; we now only refresh the material list/model change notification

adriengivry (approved):

PR #669, opened 2026-04-08
Scripts can now be saved ANYWHERE!

### LINKED ISSUE #385: Move scripts to the `Assets/` folder

adriengivry:
## Description

Previously, Overload always created a `Scripts/` folder inside of a user's project folder.
Scripts can now be located anywhere inside of the project's `Assets/` folder.
This allow scripts to be placed anywhere, including inside of sub-folders.

Technically this would be a breaking change, but a nice window has been added to automatically relocate scripts from `Scripts/` to `Assets/Scripts/` when an old project is opened.

image

In Lua, behaviours can be retrieved using `GetBehaviour`:
```lua
--- Assuming a script Foo.lua is located in a Scripts/ folder
--- The following calls are valid:
self.owner:GetBehaviour("Foo") --- by script name
self.owner:GetBehaviour("Scripts/Foo") --- by script path
self.owner:GetBehaviour("Scripts/Foo.lua") --- by script path with extension
```

## Related Issue(s)

Fixes #385

## Review Guidance

Write here.

## Screenshots/GIFs

image

## Checklist

- [x] I have updated the documentation accordingly

### DIFF
--- a/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaActorBindings.cpp
+++ b/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaActorBindings.cpp
@@ -76,7 +76,7 @@ void BindLuaActor(sol::state& p_luaState)
 			OvCore::ECS::Components::Behaviour* behaviour = nullptr;
 			for (auto& [key, b] : p_this.GetBehaviours())
 			{
-				if (b.GetScriptName() == p_name)
+				if (std::filesystem::path(b.name).stem().string() == p_name)
 				{
 					behaviour = &b;
 					break;
--- a/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaActorBindings.cpp
+++ b/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaActorBindings.cpp
@@ -4,6 +4,8 @@
 * @licence: MIT
 */
 
+#include <filesystem>
+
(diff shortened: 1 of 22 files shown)


Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaActorBindings.cpp
```diff
@@ -70,7 +72,29 @@ void BindLuaActor(sol::state& p_luaState)
...
+			{
+				behaviour = p_this.GetBehaviour(p_name);
+			}
+
+			if (!behaviour && std::filesystem::path(p_name).extension().empty())
+			{
+				behaviour = p_this.GetBehaviour(p_name + ".lua");
+			}
+
+			if (behaviour)
```
adriengivry, line 97: I'm not exactly thrilled about this, but this is necessary to keep all the `GetBehaviour("{name}")` from breaking after this update. Another option would be to move this code to `GetBehaviour` itself, but having lua-specific code in there wouldn't make much sense. Could definitely be improved in the future tho.

PR #672, opened 2026-04-08
Fixed selection outline reverting to T-pose when clearing skinned mesh materials

### LINKED ISSUE #671: Selection outline reverts to T-Pose when clearing material on Skinned Mesh Renderer
## Description

When adding a **Skinned Mesh Renderer** component to an Actor and clearing one of its assigned materials, the selection outline in the Scene View no longer follows the animated mesh. Instead, it reverts to displaying the Actor in its default T-Pose.

## To Reproduce

1. Create or select an Actor with an animated skinned mesh
2. Add a **Skinned Mesh Renderer** component (if not already present)
3. Assign materials to the mesh (if not already set to `Default`)
4. Clear (remove) one of the materials from the component
5. Observe the selection outline in the Scene View

## Expected behavior

The selection outline should continue to follow the animated pose of the Actor, matching the deformation of the skinned mesh in real time.

## Screenshots

Image

Gopmyc:
## Description
This PR fixes a bug where the Scene View selection outline for skinned meshes reverts to the T-pose after clearing a material slot in `CMaterialRenderer`.

Root cause:
- Outline skinning was enabled only when the original material existed and supported skinning.
- Clearing a material sets that slot to `nullptr`, which disabled skinning for outline rendering.

Fix:
- In `OutlineRenderFeature`, skinning is now decided from the resolved target material (regular material with `OUTLINE_PASS` or fallback outline material), not only the original material pointer.
- Added `mesh->HasSkinningData()` checks to keep skinning activation consistent and safe in both stencil and outline paths.

## Related Issue(s)
Fixes #671

## Review Guidance
1. Select an actor with an animated skinned mesh.
2. Verify the selection outline follows the animated pose.
3. Clear one material slot from the actor's `Material Renderer`.
4. Verify the outline still follows animation and no longer reverts to T-pose.

## Checklist
- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~

### DIFF
--- a/Sources/OvEditor/src/OvEditor/Rendering/OutlineRenderFeature.cpp
+++ b/Sources/OvEditor/src/OvEditor/Rendering/OutlineRenderFeature.cpp
@@ -19,6 +19,7 @@ namespace
 	constexpr uint32_t kStencilMask = 0xFF;
 	constexpr int32_t kStencilReference = 1;
 	constexpr std::string_view kOutlinePassName = "OUTLINE_PASS";
+	const std::string kSkinningFeatureName = std::string{ OvCore::Rendering::SkinningUtils::kFeatureName };
 
 	using MaterialList = OvCore::ECS::Components::CMaterialRenderer::MaterialList;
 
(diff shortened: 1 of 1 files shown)


adriengivry (changes requested):
Looks mostly good, I haven't tested it, but I'm guessing if a material uses a shader that doesn't support skinning, this will cause the outline to still use skinning, which shouldn't be the case. The original material needs to be considered.

Can you confirm that this bug occurs?

To repro the potential bug:
1. Duplicate the standard shader
2. Remove skinning feature from the duplicated shader
3. Create a material using the non-skinned standard shader
4. Create an Actor with a model renderer, material renderer, and skinned mesh renderer.
5. Apply the new material to the material renderer
6. Set an animation
7. Notice that while the model isn't skinned, the outline is

Gopmyc:
> Looks mostly good, I haven't tested it, but I'm guessing if a material uses a shader that doesn't support skinning, this will cause the outline to still use skinning, which shouldn't be the case. The original material needs to be considered.
> 
> Can you confirm that this bug occurs?
> 
> To repro the potential bug:
> 
> 1. Duplicate the standard shader
> 2. Remove skinning feature from the duplicated shader
> 3. Create a material using the non-skinned standard shader
> 4. Create an Actor with a model renderer, material renderer, and skinned mesh renderer.
> 5. Apply the new material to the material renderer
> 6. Set an animation
> 7. Notice that while the model isn't skinned, the outline is

Good catch, confirmed and fixed !

adriengivry (approved):
LGTM!

PR #701, opened 2026-04-14
Prevent crash when dragging a folder into its own child

### LINKED ISSUE #699: Drag-and-dropping a parent folder onto one of its children crashes
## Description
Overload crashes when a parent folder is being dragged and dropped onto one of its children folder.

## To Reproduce
1. Create folder "A"
2. Create folder "B" as a child of "A"
3. Drag and drop "A" onto "B"
4. Notice crash

## Expected behavior
Similarly to what we do in the hierarchy, we should prevent such behaviour.

Gopmyc:
## Description
This PR fixes a crash in the Asset Browser when a folder is dragged and dropped into one of its own descendants.

The fix adds a local guard in the folder drag-and-drop handler to detect invalid parent->child (or self) moves before filesystem rename/copy operations are executed.  
When such a move is detected, the operation is safely ignored and a warning is logged.

## Related Issue(s)
Fixes #699

## Review Guidance
Please focus review on:
- Folder DDTarget logic in `AssetBrowser.cpp`
- The new path ancestry guard used before move/copy operations
- Regression risk on valid drag/drop scenarios (project/project and engine/project)

## Screenshots/GIFs
N/A (behavioral crash fix; no UI change)

## Checklist

### DIFF
--- a/Sources/OvEditor/src/OvEditor/Panels/AssetBrowser.cpp
+++ b/Sources/OvEditor/src/OvEditor/Panels/AssetBrowser.cpp
@@ -130,27 +130,8 @@ namespace
 
 	bool IsPathSameOrDescendant(const std::filesystem::path& p_path, const std::filesystem::path& p_ancestor)
 	{
-		const std::filesystem::path ancestor = p_ancestor.lexically_normal();
-		std::filesystem::path current = p_path.lexically_normal();
-
-		while (!current.empty())
-		{
-			if (current == ancestor)
-			{
-				return true;
-			}
-
-			const std::filesystem::path parent = current.parent_path();
-
-			if (parent == current)
-			{
-				break;
-			}
-
-			current = parent;
-		}
-
-		return false;
+		const std::filesystem::path relativePath = p_path.lexically_normal().lexically_relative(p_ancestor.lexically_normal());
+		return !relativePath.empty() && *relativePath.begin() != "..";
 	}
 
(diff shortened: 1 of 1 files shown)


Sources/OvEditor/src/OvEditor/Panels/AssetBrowser.cpp
```diff
@@ -128,6 +128,31 @@ namespace
...
+			if (parent == current)
+			{
+				break;
+			}
+
+			current = parent;
+		}
+
+		return false;
+	}
```
adriengivry, line 135: While this works, we could leverage `std::filesystem` power:
```cpp
bool IsPathSameOrDescendant(const std::filesystem::path& p_path, const std::filesystem::path& p_ancestor)
{
	auto rel = p_path.lexically_normal().lexically_relative(p_ancestor.lexically_normal());
	return !rel.empty() && *rel.begin() != "..";
}
```
Make sure it works on Windows, I only tested it on Linux.

adriengivry (changes requested):

Sources/OvEditor/src/OvEditor/Panels/AssetBrowser.cpp
Gopmyc: Good catch !
I switched it to use lexically_relative as suggested

adriengivry (approved):

PR #714, opened 2026-04-16
Add minimal prefab workflow based on actor XML serialization

### LINKED ISSUE #25: Implement a prefab system
> [!NOTE]
> This issue is not open for PRs at the moment. We're looking for design proposals to introduce a prefab system as smoothly as possible. See the description for our initial thoughts.

## Description
Game projects frequently reuse parts of a scene across multiple scenes. To support this workflow, we need a prefab system that allows actors to be shared and reused.

### Option A — Dedicated Prefab File Format (`.ovprefab`)
Introduce a new file type specifically for prefabs.
**Pros:** Purpose-built for reuse; clear separation of concerns.
**Cons:** Requires building and maintaining a new asset pipeline from scratch.

### Option B — Scene-referencing Scenes
Allow scenes to embed references to other scenes, making scenes composable units rather than flat actor lists.
**Pros:** Leverages the existing scene system; no new file format needed.
**Cons:**
- The scene node structure would need to support a new `scene reference` node type.
- The hierarchy view would need to clearly distinguish scene references from local actors, so it's obvious to the user that referenced content may change when the source file changes.
- Overriding content within a scene reference (e.g. changing a property for a specific instance) would be difficult or impossible without additional complexity.

## Synchronization Problem

Both approaches share a critical issue: **keeping prefab instances in sync with their source**. A previous PR added the ability to create prefab assets, but without any synchronization logic to propagate changes from the source file to existing instances. Without this, the system provides limited value over simple copy-paste.

Any viable design proposal must address how updates to a source prefab or scene are reflected across all references to it.

Gopmyc:
## Description
This PR implements a minimal prefab workflow based on the existing actor/scene XML serialization system.

What is included:
- Added `.ovprefab` file type support in `PathParser`.
- Added editor integration to save an actor hierarchy as a prefab from the Hierarchy contextual menu.
- Added prefab instantiation by drag-and-drop in Scene View.
- Reused `Actor::OnSerialize` / `Actor::OnDeserialize` for prefab save/load to stay consistent with `.ovscene` behavior.
- Added prefab-aware asset retargeting propagation for renames (scripts, models, materials, sounds) so references in prefab files stay valid.
- Reduced log noise during prefab instantiation:
  - no per-actor `Actor created` spam
  - keep a single log entry:
    `Prefab instantiated from disk: <path>`

## Related Issue(s)
Fixes #25

## Review Guidance
Key areas to review:
- Prefab file type integration:
  - `Sources/OvTools/include/OvTools/Utils/PathParser.h`
  - `Sources/OvTools/src/OvTools/Utils/PathParser.cpp`
- Prefab save/load actions:
  - `Sources/OvEditor/include/OvEditor/Core/EditorActions.h`
  - `Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp`
- Editor UI integration:
  - `Sources/OvEditor/src/OvEditor/Panels/Hierarchy.cpp` (Save as prefab)
  - `Sources/OvEditor/include/OvEditor/Panels/SceneView.h`
  - `Sources/OvEditor/src/OvEditor/Panels/SceneView.cpp` (drop instantiate)
- Minor UI/type support:
  - `Sources/OvCore/src/OvCore/Helpers/GUIHelpers.cpp`
  - `Sources/OvEditor/src/OvEditor/Core/EditorResources.cpp`

Suggested manual checks:
1. Create an actor with children/components, then use **Save as prefab...**.
2. Drag the `.ovprefab` into Scene View and verify hierarchy/component restoration.
3. Confirm log output only shows the prefab instantiation message (no actor creation spam).
4. Rename referenced assets/scripts and verify `.ovprefab` references are retargeted.

## Screenshots/GIFs
N/A

## Checklist

### DIFF
--- a/Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
+++ b/Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
@@ -53,6 +53,25 @@
 namespace
 {
 	constexpr std::string_view kDefaultMaterialPath = ":Materials\\Default.ovmat";
+	bool s_logActorCreation = true;
+
+	class ScopedActorCreationLog
+	{
+	public:
+		explicit ScopedActorCreationLog(bool p_enabled) :
+			m_previousState(s_logActorCreation)
+		{
+			s_logActorCreation = p_enabled;
+		}
+
+		~ScopedActorCreationLog()
+		{
+			s_logActorCreation = m_previousState;
+		}
+
+	private:
+		bool m_previousState;
+	};
 
 	void SerializeActorHierarchy(
 		tinyxml2::XMLDocument& p_doc,
@@ -723,7 +742,8 @@ OvCore::ECS::Actor & OvEditor::Core::EditorActions::CreateEmptyActor(bool p_focu
 	if (p_focusOnCreation)
 		SelectActor(instance);
 
-	OVLOG_INFO("Actor created");
+	if (s_logActorCreation)
+		OVLOG_INFO("Actor created");
 
 	return instance;
 }
@@ -904,6 +924,7 @@ OvCore::ECS::Actor* OvEditor::Core::EditorActions::InstantiatePrefabFromDisk(
 
 	std::vector<OvCore::ECS::Actor*> createdActors;
 	std::unordered_map<int64_t, OvCore::ECS::Actor*> actorLookup;
+	ScopedActorCreationLog actorCreationLogScope(false);
 
 	for (tinyxml2::XMLElement* currentActor = actorsRoot->FirstChildElement("actor"); currentActor; currentActor = currentActor->NextSiblingElement("actor"))
 	{
--- a/Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
+++ b/Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
@@ -12,6 +12,7 @@
 #include <iostream>
 #include <ranges>
 #include <string_view>
+#include <unordered_map>
 #include <vector>
 #include <tinyxml2.h>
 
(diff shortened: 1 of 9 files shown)


Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -708,7 +742,8 @@ OvCore::ECS::Actor & OvEditor::Core::EditorActions::CreateEmptyActor(bool p_focu
 	if (p_focusOnCreation)
 		SelectActor(instance);
 
-	OVLOG_INFO("Actor created");
+	if (s_logActorCreation)
+		OVLOG_INFO("Actor created");
```
adriengivry: To be honest, I don't think this log is even needed anymore. Let's get rid of it. This way, we won't even have to add any workaround to disable it for prefabs

adriengivry (commented):
With this implementation, is it possible to edit a prefab, and see its edits affect all the scenes it is included in?

I was thinking, another way to go about making prefab (that would be even simpler) would be to allow scenes to reference other scenes. Essentially, scenes would be prefabs. This would completely remove the concept of "prefabs", but would allow nested scenes, and make editing subscene super easy.

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: Good point, totally agree

Gopmyc:
> With this implementation, is it possible to edit a prefab, and see its edits affect all the scenes it is included in?
> 
> I was thinking, another way to go about making prefab (that would be even simpler) would be to allow scenes to reference other scenes. Essentially, scenes would be prefabs. This would completely remove the concept of "prefabs", but would allow nested scenes, and make editing subscene super easy.

Great point, and I agree.

With the current implementation, prefab instances are serialized snapshots, so editing the prefab file does not automatically propagate to existing scene instances yet.

Your “scene references scene” idea makes a lot of sense as a long-term direction. It would unify the workflow and make subscene/prefab editing much cleaner.

For this PR, I intentionally kept the scope minimal to match issue #25, but I’m happy to open a follow-up issue to explore nested scenes as the next step.

adriengivry:
There is no point in merging this one and adding a new asset type with a bunch of logic associated if we end up simplifying it later by introducing scene references.

I'd rather go with the simpler scene reference implementation than integrating a temporary feature like this one.

Gopmyc:
> There is no point in merging this one and adding a new asset type with a bunch of logic associated if we end up simplifying it later by introducing scene references.
> 
> I'd rather go with the simpler scene reference implementation than integrating a temporary feature like this one.

Makes sense, I agree.
If the target direction is scene references, merging a temporary prefab asset type would just add churn.
I’ll stop this approach and pivot to a minimal scene-reference implementation instead.
I’ll open a follow-up PR focused on nested scene references so we can review the final direction directly.

Issue #719, opened 2026-04-16
Metallic property not properly loaded with embedded materials


adriengivry:
## Description
The metallic property is always 0, even when the software used to author the model set a different value.

## To Reproduce
1. Open Blender
2. Change the metallic value of the default cube
3. Export to FBX
4. Import into Overload
5. Add model into a scene
6. Slick on the created actor
7. Double click on the material in the inspector
8. In the material editor, find the "Metallic" property
9. Notice that it's set to 0 instead of the value set in Blender

## Expected behavior
The metallic property should import properly.

## Screenshots

<img width="1717" height="1066" alt="Image" src="https://github.com/user-attachments/assets/3ea24371-0640-463d-940f-6b2689caaae8" />

<img width="1541" height="977" alt="Image" src="https://github.com/user-attachments/assets/965f487c-0962-41be-bb96-8a68df1d254a" />

adriengivry:
I'd be curious to see if we get the same issue with glTF models.

adriengivry:
Fixed by https://github.com/Overload-Technologies/Overload/pull/721

PR #721, opened 2026-04-17
Fix FBX metallic import and inspector refresh for embedded materials on model reload

Gopmyc:
## Description
This PR fixes FBX metallic import/reload behavior and improves editor refresh flow for embedded materials.

Changes included:
- Add a fallback in Assimp material parsing to read metallic from `AI_MATKEY_REFLECTIVITY` when `AI_MATKEY_METALLIC_FACTOR` is missing (common with FBX exports).
- Normalize model paths when reloading embedded resources so embedded materials/textures are reliably matched and refreshed.
- Refresh `Inspector` and `Material Editor` panels after model reload actions (Asset Browser reload and Asset Properties apply), so updated embedded material values are visible immediately without closing/reopening panels.

## Related Issue(s)
#719

## Review Guidance
1. Import an FBX model with metallic set to `0` in Blender and open one of its embedded materials.
2. Change metallic to `1` in Blender, re-export over the same FBX file.
3. In Overload, reload the model:
   - either from Asset Browser context menu (`Reload`)
   - or from Asset Properties (`Apply`) for the model.
4. Verify the embedded material metallic value updates immediately in the Material Editor/Inspector, without restarting the editor.

## Screenshots/GIFs
N/A (resource reload and inspector refresh behavior fix)

## Checklist

### DIFF
--- a/Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
+++ b/Sources/OvCore/src/OvCore/ResourceManagement/ModelManager.cpp
@@ -11,6 +11,7 @@
 #include <OvCore/ResourceManagement/TextureManager.h>
 #include <OvRendering/Resources/Parsers/EmbeddedAssetPath.h>
 #include <OvTools/Filesystem/IniFile.h>
+#include <OvTools/Utils/PathParser.h>
 
 namespace
 {
@@ -69,11 +70,16 @@ namespace
 		TIndexParser p_parseIndex
 	)
 	{
+		const std::string normalizedModelPath = OvTools::Utils::PathParser::MakeNonWindowsStyle(p_modelPath);
+
 		for (auto& [resourcePath, resource] : p_resourceManager.GetResources())
 		{
 			(void)resource;
 			const auto embeddedAssetPath = OvRendering::Resources::Parsers::ParseEmbeddedAssetPath(resourcePath.string());
-			if (!embeddedAssetPath || embeddedAssetPath->modelPath != p_modelPath)
+			if (
+				!embeddedAssetPath ||
+				OvTools::Utils::PathParser::MakeNonWindowsStyle(embeddedAssetPath->modelPath) != normalizedModelPath
+			)
 			{
 				continue;
 			}
--- a/Sources/OvEditor/src/OvEditor/Panels/AssetBrowser.cpp
+++ b/Sources/OvEditor/src/OvEditor/Panels/AssetBrowser.cpp
@@ -25,6 +25,7 @@
 #include <OvEditor/Core/EditorResources.h>
 #include <OvEditor/Panels/AssetBrowser.h>
 #include <OvEditor/Panels/AssetProperties.h>
+#include <OvEditor/Panels/Inspector.h>
 #include <OvEditor/Panels/MaterialEditor.h>
 #include <OvEditor/Settings/EditorSettings.h>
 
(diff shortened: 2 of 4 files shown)


adriengivry:
That's interesting...

Could you share some documentation from Assimp about REFLECTIVITY being used in place of METALLIC_FACTOR? Should it be mapped 1 to 1? Are there cases where both could be defined? Is this behaviour shared by other common exporters (Maya, 3DS Max) or is it specific to Blender?

Gopmyc:
> That's interesting...
> 
> Could you share some documentation from Assimp about REFLECTIVITY being used in place of METALLIC_FACTOR? Should it be mapped 1 to 1? Are there cases where both could be defined? Is this behaviour shared by other common exporters (Maya, 3DS Max) or is it specific to Blender?

I took a deeper look at Assimp’s FBX/material code to make sure this fallback is justified.

From Assimp’s side:
- `AI_MATKEY_METALLIC_FACTOR` is the official PBR metallic scalar key (`material.h`).
- `AI_MATKEY_REFLECTIVITY` is a separate legacy key.
- In the FBX importer, `ReflectionFactor` is written to `AI_MATKEY_REFLECTIVITY`.
- `AI_MATKEY_METALLIC_FACTOR` is only written when Assimp finds specific PBR properties (for example `Maya|metallic`).

So the current change does **not** replace metallic with reflectivity globally.  
It only uses reflectivity as a fallback when metallic is missing, and if both are present, metallic still wins.

To your questions directly:
1. Is it always 1:1?  
Not in a strict semantic sense. This is a practical fallback for FBX cases where Assimp does not expose a metallic scalar through `AI_MATKEY_METALLIC_FACTOR`.

2. Can both exist?  
Yes. In that case we already prioritize `AI_MATKEY_METALLIC_FACTOR`.

3. Blender-only or shared?  
Not Blender-only. This behavior comes from Assimp’s FBX conversion path. Blender is one reproducible case, but the importer logic is shared. Maya has an explicit metallic scalar path, and 3ds Max has explicit metalness texture handling.

If you prefer, I can narrow the fallback so it applies only for `.fbx` imports to reduce any ambiguity outside FBX ?

adriengivry (changes requested):
Thanks for the clarification.

I tested it with a simple cube and it seems to now load the correct metallic and roughness values. However I noticed that when:
- a metallic texture is being used
- `AI_MATKEY_METALLIC_FACTOR` doesn't return `AI_SUCCESS`
- `AI_MATKEY_REFLECTIVITY` return `AI_SUCCESS` with a `0` value
The metallic factor will be 0, and will prevent the metallic texture from having any effect.

We should probably treat metallic and reflectivity separately, and only apply override the metallic factor to use reflectivity when:
- the metallic is zero
- AND the reflectivity is non-zero
- AND there is no metallic texture attached

Reflectivity would essentially be a fallback if no metallic factor NOR metallic texture is used.

Gopmyc:
> Thanks for the clarification.
> 
> I tested it with a simple cube and it seems to now load the correct metallic and roughness values. However I noticed that when:
> 
> * a metallic texture is being used
> * `AI_MATKEY_METALLIC_FACTOR` doesn't return `AI_SUCCESS`
> * `AI_MATKEY_REFLECTIVITY` return `AI_SUCCESS` with a `0` value
>   The metallic factor will be 0, and will prevent the metallic texture from having any effect.
> 
> We should probably treat metallic and reflectivity separately, and only apply override the metallic factor to use reflectivity when:
> 
> * the metallic is zero
> * AND the reflectivity is non-zero
> * AND there is no metallic texture attached
> 
> Reflectivity would essentially be a fallback if no metallic factor NOR metallic texture is used.

Great catch, thanks for testing this edge case.

You’re absolutely right: using reflectivity too early could force metallic to 0 and cancel the metallic map contribution.

I updated the logic so reflectivity is now a strict fallback only when:
- `AI_MATKEY_METALLIC_FACTOR` is not available
- no metallic texture is attached
- and reflectivity is non-zero

Priority is now:
1) metallic factor
2) metallic texture fallback (`metallic = 1.0`)
3) reflectivity fallback (last resort only)

This keeps metallic and reflectivity separate and avoids overriding valid metallic-map workflows.

adriengivry (approved):
LGTM!

PR #723, opened 2026-04-17
Fix material editor widgets not updating after shader compilation from asset browser

### LINKED ISSUE #471: Material editor shader settings & features not updated after compilation
**Description**
The material editor currently generate widgets for shader uniforms and features when a material is selected, or its associated shader updated. Compiling a shader doesn't trigger the material editor to update its widgets.

**To Reproduce**
1. Open a material in the material editor
2. Add a feature to the shader used by this material (example: add `#feature FOO` somewhere in the shader code)
3. Compile the shader
4. Notice that the material editor isn't updated to show this new feature

**Expected behavior**
The material editor should always be updated after a shader has been compiled.

**Screenshots**
_N/A_

Gopmyc:
## Description
The material editor currently regenerates shader uniforms/features when selecting a material or changing its shader, but not when compiling that shader from the Asset Browser contextual menu.

This PR updates the shader contextual compile flow to:
- keep a reference to the shader that was just compiled/reloaded
- refresh the Material Editor only when its current target material uses that compiled shader
- force a property update before refresh so newly added uniforms/features appear immediately

This fixes the missing UI update while keeping the behavior scoped and minimal.

## Related Issue(s)
Fixes #471

## Review Guidance
1. Open a material in Material Editor.
2. Add a new feature in its shader (e.g. `#feature FOO`).
3. Compile the shader from the Asset Browser contextual menu.
4. Verify the Material Editor updates immediately (features/properties regenerated).

## Screenshots/GIFs
N/A

## Checklist

### DIFF
--- a/Sources/OvEditor/include/OvEditor/Core/EditorActions.h
+++ b/Sources/OvEditor/include/OvEditor/Core/EditorActions.h
@@ -6,6 +6,8 @@
 
 #pragma once
 
+#include <filesystem>
+
 #include <OvCore/Global/ServiceLocator.h>
 #include <OvEditor/Core/Context.h>
 #include <OvEditor/Core/PanelsManager.h>
@@ -264,6 +266,11 @@ namespace OvEditor::Core
 		* Compile the given shader
 		*/
 		void CompileShader(OvRendering::Resources::Shader& p_shader);
+
+		/**
+		* Compile the shader at the given resource path
+		*/
+		void CompileShader(const std::filesystem::path& p_shaderPath);
 
 		/**
 		* Save every materials to their respective files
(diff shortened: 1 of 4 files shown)


adriengivry (changes requested):
This indeed fixes it for the inspector, but still doesn't work with the action in the menu bar:

Image

We should find a proper fix that would apply to all code-paths. Having a single function in `EditorActions` that can properly handle compilation and UI refresh would be great. This function should be used everywhere (AssetBrowser, MenuBar, MaterialEditor, etc...).

Gopmyc:
> This indeed fixes it for the inspector, but still doesn't work with the action in the menu bar:
> 
> Image
> We should find a proper fix that would apply to all code-paths. Having a single function in `EditorActions` that can properly handle compilation and UI refresh would be great. This function should be used everywhere (AssetBrowser, MenuBar, MaterialEditor, etc...).

Nice catch ! It’s fixed

adriengivry (approved):
LGTM!

PR #729, opened 2026-04-17
Fix Crash when deleting a folder outside of Overload and trying to expand it afterwards#725

### LINKED ISSUE #725: Crash when deleting a folder outside of Overload and trying to expand it afterwards
## Description
Crash when deleting a folder outside of Overload and trying to expand it afterwards.

## To Reproduce
1. Create a folder in Overload
2. Collapse the folder
3. Delete the folder outside of Overload
4. Try expanding the folder in Overload
5. Notice the crash

## Expected behavior
Shouldn't crash

Synyproxy:
## Description
Fixes this bug:
Delete a folder outside of Overload, then try expanding the folder in Overload causes crash.
Now expanding the folder shows error message.
Also fixed deleting that folder from the asset browser.

## Related Issue(s)
https://github.com/Overload-Technologies/Overload/issues/725
Fixes #725

## Review Guidance
When expanding a delete folder, user get's feedback message.
Also fixed deleting that folder from the asset browser

## Checklist

- [x ] My code follows the project's code style guidelines
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] I have updated the documentation accordingly
- [ x] My changes don't generate new warnings or errors


adriengivry (approved):
~~LGTM!~~

adriengivry (changes requested):
Actually contextual actions on the missing folder cause crash (show in explorer, import here, rename...)

Image

Synyproxy (commented):
this should handle the other asset browser context menu actions on deleted folder

Sources/OvEditor/src/OvEditor/Panels/AssetBrowser.cpp
```diff
@@ -257,6 +292,12 @@ namespace
 
 		void CreateNewShader(const std::string& p_shaderName, std::optional<const std::string_view> p_type)
 		{
+			if (!ValidateFolderPath(filePath))
+			{
+				ShowPathDeletedError("Create shader");
+				return;
+			}
```
adriengivry: `ValidateFolderPath` could automatically show the error message when it fails.
So we can have something like:
```cpp
if (!ValidateFolderPath(filePath, "Create shader") return;
```

Sources/OvEditor/src/OvEditor/Panels/AssetBrowser.cpp
```diff
@@ -135,6 +136,27 @@ namespace
...
+			MessageBox errorMessage(
+				"Action failed",
+				"Cannot perform this action because the target folder was deleted externally",
+				MessageBox::EMessageType::ERROR,
+				MessageBox::EButtonLayout::OK
+			);
+			return true;
+		}
+		return false;
+	}
```
adriengivry: Probably doesn't need a return value, and could simply log to the console with `OVLOG_ERROR`

adriengivry (changes requested):

Gopmyc:
> this should handle the other asset browser context menu actions on deleted folder

Hi !
Do you need help with this PR ?

adriengivry (approved):

PR #750, opened 2026-04-20
Add project window icon setting and apply it to runtime + Windows built executable icon

### LINKED ISSUE #90: Capability to add your own icon to your game executable.
Adding the capability to add your own logo to an executable is a common feature in a game engine. This should be fairly simple to implement.

### LINKED ISSUE #755: Overload engine process icon is vertically flipped (default icon)
## Description
On Windows, the default process icon of the Overload engine executable appears vertically flipped (upside down).

## To Reproduce
1. Build Overload on Windows.
2. Launch the engine executable (`OvGame.exe`) from the repository build output.
3. Observe the flipped process icon in the top left corner of the window.

## Expected behavior
The default Overload engine process icon should have the same orientation as the original icon asset (not vertically flipped).

Gopmyc:
## Description
This PR introduces project-level window icon support and applies it in two places:

1. Runtime window icon:
- Adds `window_icon` to project settings
- Uses the configured icon at game runtime when available
- Keeps the existing fallback icon behavior when no icon is configured or found

2. Windows build executable icon:
- During build export, updates the generated game `.exe` icon resources using the configured `window_icon`
- If no icon is configured (or file is missing), keeps the default executable icon behavior

## Related Issue(s)
Fixes #90
Fixes #755 

## Review Guidance
Please review:
- project settings addition and integrity checks for `window_icon`
- runtime icon resolution/fallback behavior
- Windows-only executable icon resource update in build pipeline

## AI Usage Disclosure
Refactored code / Generated new code / Debugging

## Checklist
- [ ] ~~When applicable, I have commented my code, particularly in hard-to-understand areas~~
- [ ] ~~When applicable, I have updated the documentation accordingly~~


Sources/OvGame/src/OvGame/Core/Context.cpp
```diff
@@ -93,8 +94,27 @@ OvGame::Core::Context::Context() :
...
-	window->SetIconFromMemory(reinterpret_cast<uint8_t*>(iconRaw.data()), 16, 16);
+	bool useFallbackWindowIcon = true;
+	const std::string windowIconPath = projectSettings.GetOrDefault("window_icon", std::string{});
+	if (!windowIconPath.empty())
+	{
+		const std::string normalizedPath = OvTools::Utils::PathParser::MakeNonWindowsStyle(windowIconPath);
+		const std::filesystem::path realPath = normalizedPath.starts_with(':') ?
+			(engineAssetsPath / normalizedPath.substr(1)).lexically_normal() :
+			(projectAssetsPath / normalizedPath).lexically_normal();
```
adriengivry: This is duplicated from `GetRealPath`. In our previous conversations we decided to move `GetRealPath` implementation to `OvTools` and update duplicated `GetRealPath` definitions to use the new one in `OvTool`. This should follow the same logic.

Sources/OvTools/include/OvTools/Utils/SystemCalls.h
```diff
@@ -39,6 +40,13 @@ namespace OvTools::Utils
 		*/
 		static void RunProgram(const std::string& p_file, const std::string& p_workingDir = "");
 
+		/**
+		* On Windows, apply the given icon image to an executable icon resource.
+		* @param p_executablePath
+		* @param p_iconPath
+		*/
+		static bool SetWindowsExecutableIcon(const std::filesystem::path& p_executablePath, const std::filesystem::path& p_iconPath);
```
adriengivry: The name of this function shouldn't contain the platform (Windows). It's up to the implementation to decide what to do when it's called. The description could be made more generic and a note could be added:

```cpp
/**
* Apply the given icon image to an executable icon resource.
* @note No-op on Linux, as executable files don't embed icons 
*/
static bool SetExecutableIcon(...);
```

Sources/OvTools/src/OvTools/Utils/SystemCalls.cpp
```diff
@@ -10,6 +10,10 @@
 #ifdef _WIN32
 #include <Windows.h>
 #include <ShlObj.h>
+#include <gdiplus.h>
+#include <objidl.h>
+#pragma comment(lib, "gdiplus.lib")
+#pragma comment(lib, "ole32.lib")
```
adriengivry: These should be added to [`OvTools/premake5.lua`](https://github.com/Overload-Technologies/Overload/blob/main/Sources/OvTools/premake5.lua) instead.

It should look something like that (not tested):
```lua
filter { "system:windows" }
  links {
    "gdiplus.lib",
    "ole32.lib",
  }
filter {}
```

Sources/OvTools/src/OvTools/Utils/SystemCalls.cpp
```diff
@@ -80,6 +85,121 @@ void OvTools::Utils::SystemCalls::RunProgram(const std::string& p_file, const st
 #endif
 }
 
+bool OvTools::Utils::SystemCalls::SetWindowsExecutableIcon(const std::filesystem::path& p_executablePath, const std::filesystem::path& p_iconPath)
```
adriengivry: Use [`[[maybe_unused]]`](https://en.cppreference.com/cpp/language/attributes/maybe_unused) for arguments that might not be used by all implementations (e.g. `p_executablePath` and `p_iconPath`, this way there is no need for:
```cpp
(void)p_executablePath;
(void)p_iconPath;
```

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -472,6 +472,19 @@ void OvEditor::Core::EditorActions::BuildAtLocation(
...
+							if (!windowIconPath.empty())
+							{
+								const std::filesystem::path windowIconRealPath{ GetRealPath(windowIconPath) };
+
+								if (std::filesystem::exists(windowIconRealPath))
+								{
+									OvTools::Utils::SystemCalls::SetWindowsExecutableIcon(p_buildPath / executableName, windowIconRealPath);
+								}
+							}
+#endif
```
adriengivry, line 528: This shouldn't be in a conditional compile block, as `SetWindowsExecutableIcon` (or `SetExecutableIcon` after we rename it), will internally handle different behaviours on different platforms.

If we decide to add an implementation for a specific platform, this code-path will benefit from this change immediately.

Sources/OvTools/src/OvTools/Utils/SystemCalls.cpp
```diff
@@ -80,6 +85,121 @@ void OvTools::Utils::SystemCalls::RunProgram(const std::string& p_file, const st
...
+		{
+			EndUpdateResourceW(updateHandle, TRUE);
+			return false;
+		}
+
+		return EndUpdateResourceW(updateHandle, FALSE) != FALSE;
+	}();
+
+	Gdiplus::GdiplusShutdown(gdiplusToken);
+	return success;
```
adriengivry: From my testing, this returns `false` on Windows with TGAs, and `true` with PNGs, yet the executable icon is never replaced.

Image

_Using `lion.tga` from Showroom (`success = false`)_

Image

_Using `gold-scuffed_basecolor.png` from Showroom (`success = true`)_

adriengivry (changes requested):
Executable icon replacement doesn't seem to work, and a few additional changes.

Sources/OvGame/src/OvGame/Core/Context.cpp
Gopmyc: Yes, absolutely, I forgot to implement it 😅
It's fixed !

Sources/OvTools/include/OvTools/Utils/SystemCalls.h
Gopmyc: It's done !

Sources/OvTools/src/OvTools/Utils/SystemCalls.cpp
Gopmyc: Good point !
I wasn't familiar with `[[may_unused]]`

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: It's fixed

Sources/OvTools/src/OvTools/Utils/SystemCalls.cpp
Gopmyc: Okay, it seems to be sorted, but it still needs to be tested

Sources/OvTools/src/OvTools/Utils/SystemCalls.cpp
Gopmyc: Done !

Sources/OvGame/src/OvGame/Core/Context.cpp
```diff
@@ -93,8 +94,28 @@ OvGame::Core::Context::Context() :
...
+	{
+		const std::filesystem::path realPath = OvTools::Utils::PathParser::GetRealPath(
+			std::filesystem::path{ windowIconPath },
+			engineAssetsPath,
+			projectAssetsPath
+		);
+
+		if (std::filesystem::exists(realPath))
+		{
+			window->SetIcon(realPath.string());
```
adriengivry: We initially modified `SetIcon` to return a `bool` in case the operation failed, so that we can update `useFallbackWindowIcon` accordingly.

Sources/OvTools/include/OvTools/Utils/SystemCalls.h
```diff
@@ -39,6 +40,17 @@ namespace OvTools::Utils
...
 
+		/**
+		* Apply the given icon image to an executable icon resource.
+		* @note No-op on Linux, as executable files don't embed icons.
+		* @param p_executablePath
+		* @param p_iconPath
+		*/
+		static bool SetExecutableIcon(
+			[[maybe_unused]] const std::filesystem::path& p_executablePath,
+			[[maybe_unused]] const std::filesystem::path& p_iconPath
```
adriengivry: `[[maybe_unused]]` not needed in the header file. It's only needed for the compilation unit (.cpp).

adriengivry (changes requested):
Still not working with TGA images (e.g. `lion.tga` in Showroom).
Instead of letting `OvTools` handle the image loading, we should provide an array of pre-loaded bytes (e.g. `std::span<const uint8_t>`) and let `stb_image` do the heavy lifting. To simplify that, I've pushed this commit to main (https://github.com/Overload-Technologies/Overload/commit/fa329e024c7557324009b2a7b70fcefc6a1748ca) which exposes a new `OvRendering::Data::Image` class to the rest of the engine to load image data using `stb_image`.

When building, you could first create an `OvRendering::Data::Image`, check that it's valid, and pass its data to `SetExecutableIcon`.

Sources/OvGame/src/OvGame/Core/Context.cpp
Gopmyc: Good point, thanks.
I updated this flow: `Window::SetIcon` now returns `bool`, and fallback is now driven by the result (`useFallbackWindowIcon = !window->SetIcon(...)`)

Sources/OvTools/include/OvTools/Utils/SystemCalls.h
Gopmyc: Agreed, thanks.
I removed `[[maybe_unused]]` from the header and kept it only in the .cpp implementation

Gopmyc:
> Still not working with TGA images (e.g. `lion.tga` in Showroom). Instead of letting `OvTools` handle the image loading, we should provide an array of pre-loaded bytes (e.g. `std::span<const uint8_t>`) and let `stb_image` do the heavy lifting. To simplify that, I've pushed this commit to main ([fa329e0](https://github.com/Overload-Technologies/Overload/commit/fa329e024c7557324009b2a7b70fcefc6a1748ca)) which exposes a new `OvRendering::Data::Image` class to the rest of the engine to load image data using `stb_image`.
> 
> When building, you could first create an `OvRendering::Data::Image`, check that it's valid, and pass its data to `SetExecutableIcon`.

The problem should be fixed now.
I'm going to do a refactoring and cleanup before the next review, since we've reached 380 changes in the PR 😅

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -472,6 +473,24 @@ void OvEditor::Core::EditorActions::BuildAtLocation(
 						{
 							OVLOG_INFO("Game executable renamed to " + executableName);
 
+							if (const std::string windowIconPath = m_context.projectSettings.GetOrDefault("window_icon", std::string{}); !windowIconPath.empty())
+							{
+								if (const std::filesystem::path windowIconRealPath{ GetRealPath(windowIconPath) }; std::filesystem::exists(windowIconRealPath))
+								{
+									if (OvRendering::Data::Image iconImage{ windowIconRealPath }; iconImage && !iconImage.isHDR)
```
adriengivry: We might want to add a bunch of `OVLOG_WARNING` when:
- Icon didn't load properly
- Icon is HDR (and therefore not supported)
- Or `SetExecutableIcon` failed

Sources/OvWindowing/src/OvWindowing/Window.cpp
```diff
@@ -62,11 +62,19 @@ OvWindowing::Window::~Window()
...
 }
 
-void OvWindowing::Window::SetIcon(const std::string & p_filePath)
+bool OvWindowing::Window::SetIcon(const std::string& p_filePath)
 {
-	GLFWimage images[1];
-	images[0].pixels = stbi_load(p_filePath.c_str(), &images[0].width, &images[0].height, 0, 4);
-	glfwSetWindowIcon(m_glfwWindow, 1, images);
+	GLFWimage image{};
+	image.pixels = stbi_load(p_filePath.c_str(), &image.width, &image.height, 0, 4);
```
adriengivry: Since I added `OvRendering::Data::Image`, it might be worth it to use it here as well. That would also mean we can safely remove `dependdir .. "stb_image/include"` from `OvWindowing`'s dependencies.

adriengivry (changes requested):
A few comments, otherwise I like the state it's in!
(haven't tried running it locally tho)

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: I mostly agree
It's done !

Sources/OvWindowing/src/OvWindowing/Window.cpp
Gopmyc: It's donne too

Sources/OvEditor/premake5.lua
```diff
@@ -53,10 +53,10 @@ project "OvEditor"
 		"OvGame", -- Necessary to be built before the editor to allow building
 		"OvMaths",
 		"OvPhysics",
-		"OvRendering",
-		"OvTools",
 		"OvUI",
 		"OvWindowing",
+		"OvRendering",
+		"OvTools",
```
adriengivry, line 57: Why were these ones moved? They are alphabetized

Sources/OvWindowing/src/OvWindowing/Window.cpp
```diff
@@ -62,20 +60,28 @@ OvWindowing::Window::~Window()
...
 {
-	GLFWimage images[1];
-	images[0].pixels = p_data;
-	images[0].height = p_width;
-	images[0].width = p_height;
-	glfwSetWindowIcon(m_glfwWindow, 1, images);
+	GLFWimage image{};
+	image.pixels = p_data;
+	image.width = static_cast<int>(p_width);
+	image.height = static_cast<int>(p_height);
```
adriengivry: Could use [designated initialized](https://mropert.github.io/2026/01/15/designed_initializers/)

adriengivry (changes requested):

Sources/OvWindowing/src/OvWindowing/Window.cpp
```diff
@@ -62,20 +60,28 @@ OvWindowing::Window::~Window()
...
+	{
+		SetIconFromMemory(
+			static_cast<uint8_t*>(image.data),
+			static_cast<uint32_t>(image.width),
+			static_cast<uint32_t>(image.height)
+		);
+		return true;
+	}
+
+	return false;
```
adriengivry: While this mostly works, I've noticed that now images are flipped.

Image

This only happens with window icons, executable icons look like.

One way to mitigate this would be to add a boolean parameter to `OvRendering::Data::Image`'s constructor to flip vertically only (`true` by default).

```cpp
stbi_set_flip_vertically_on_load(true); // <-- instead of always true, make it true by default, and allow changing its value through Image's ctor. OpenGL typically requires images to be flipped, hence the default to true.
```

Gopmyc: I’ll take care of this issue right after resolving issue [#755](https://github.com/Overload-Technologies/Overload/pull/755) to make sure there’s no correlation

Sources/OvWindowing/src/OvWindowing/Window.cpp
Gopmyc: Yep, it's done !

Sources/OvEditor/premake5.lua
Gopmyc: I placed them because I had done some tests, I didn't commit them on purpose😅

Sources/OvWindowing/src/OvWindowing/Window.cpp
Gopmyc: (PS: Thank you for introducing me to [designated initialized](https://mropert.github.io/2026/01/15/designed_initializers/), I didn't know it was called that)

Sources/OvWindowing/src/OvWindowing/Window.cpp
adriengivry: I think this PR should fix it instead. Since we introduced a new way to load images, there is no point in fixing this on main.

Gopmyc: Fixed, thanks for the report !
I updated the window icon upload path to flip pixel rows before calling `glfwSetWindowIcon`, so icons are now displayed with the correct orientation on Windows.

This resolves this review comment, and I also took the opportunity to fix issue #755 in the same change

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -472,6 +473,58 @@ void OvEditor::Core::EditorActions::BuildAtLocation(
...
+									{
+										OVLOG_WARNING(
+											std::format(
+												"Failed to apply executable icon from \"{}\". Keeping default executable icon.",
+												windowIconRealPath.string()
+											)
+										);
+									}
+								}
+							}
```
adriengivry, line 527: I think you were right, keeping this in a `#ifdef` that is windows-specific will be better, as when I test this on Linux, I get a warning that the executable icon failed to apply 🥲

adriengivry (commented):
Last comment, ~~promise~~ 😝

Actually maybe not, looking for the Image update to enable/disable flip

Sources/OvWindowing/src/OvWindowing/Window.cpp
```diff
@@ -62,20 +62,46 @@ OvWindowing::Window::~Window()
...
-	images[0].height = p_width;
-	images[0].width = p_height;
-	glfwSetWindowIcon(m_glfwWindow, 1, images);
+	if (!p_data || p_width == 0u || p_height == 0u)
+	{
+		return;
+	}
+
+	const size_t rowSize = static_cast<size_t>(p_width) * 4u;
+	std::vector<uint8_t> flippedPixels(rowSize * static_cast<size_t>(p_height));
```
adriengivry: That's not how we should handle this. As I mentioned, we should update `OvRendering::Data::Image` to take a boolean parameter to enable or disable flipping (on by default). For the case of the memory data being flipped, we should reverse the data in `Context.cpp` for both the game and editor

Sources/OvGame/src/OvGame/Core/Context.cpp
```diff
@@ -93,8 +94,23 @@ OvGame::Core::Context::Context() :
...
...558797656659712ULL,17588834734687262208ULL,7284033640602212096ULL,14273902834169157632ULL,18381468269087692288ULL,6852253225049397248ULL,17732667349600245504ULL,16291515470083266560ULL,10022503688432981760ULL,11968059825861367552ULL,9733991836700645376ULL,14850363587428816640ULL,18381468271168132864ULL,16147400282007410688ULL,656430432014827520,18381468270950094848ULL,15715054717226194944ULL,72057596690306560,11823944635485519872ULL,15859169905251653376ULL,17084149004500473856ULL,8581352906816952064ULL,2527949855582584832ULL,18381468271419856896ULL,8581352907253225472ULL,252776704,1376441223417430016ULL,14994761349590357760ULL,10527190521537370112ULL,0,9806614576878321664ULL,18381468271671515136ULL,17156206598538401792ULL,6059619689256392448ULL,10166619973990488064ULL,18381468271403079424ULL,17444719549178451968ULL,420746240,870625192710242304,4906133035823863552ULL,18381468269289150464ULL,18381468271671515136ULL,18381468271671515136ULL,9950729769032620032ULL,14778305994951169792ULL,269422336,0,0,18381468268785833984ULL,8941923452686178304ULL,18381468270950094848ULL,3440842496,1233456333565402880ULL,0,0,0,11823944636091210240ULL,2383877888,16724143605745719296ULL,2316834816,0,0 });
```
adriengivry: We should get rid of this absurdity one day. It seems like we need to flip it, maybe it's a good time to get rid of it and have a proper PNG file in the engine resources?

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: Hahaha, I hate you (just kidding)
I will fix that

Sources/OvGame/src/OvGame/Core/Context.cpp
Gopmyc: Done
I replaced the hardcoded icon bytes with a proper engine PNG resource

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: I kept executable icon patching inside a Windows-only #ifdef to avoid Linux warnings

Sources/OvWindowing/src/OvWindowing/Window.cpp
Gopmyc: I added an optional flip flag to `OvRendering::Data::Image` (default on) and moved memory row reversal to `Context.cpp` for both : game and editor

adriengivry (approved):

PR #776, opened 2026-04-24
Implement `Prefab` workflow

### LINKED ISSUE #25: Implement a prefab system

Gopmyc:
## Description
This PR delivers the prefab MVP workflow in the editor end-to-end: creating prefab assets from actors, instantiating prefabs, maintaining actor-to-prefab links ("source instance"), editing prefab instances (apply/revert), and propagating asset renames through scenes and prefabs.

Implemented:
- Added dedicated `.ovprefab` file type support in asset typing/parsing.
- Added prefab icon resource and prefab picker label support.
- Added `Save as Prefab...` in actor contextual menu.
- Added prefab serialization/instantiation utilities (`PrefabOperations`).
- Added actor `prefab_source` support (API + serialization/deserialization).
- Added drag & drop prefab instantiation into scene view.
- Added `Create > From prefab...` in actor creation menu (hierarchy + menu bar).
- Added `Apply to Prefab` and `Revert to Prefab` for prefab-linked actors.
- Added hierarchy prefab visual state (prefab icon for linked instance root + children).
- Extended rename propagation logic to include `.ovprefab` files.
- Added prefab rename/delete propagation to current scene actor prefab references.
- Extended script migration retargeting to include `.ovprefab` files.

## Related Issue(s)
Fixes #25

## Review Guidance
Please validate the prefab workflow in this order:
1. Create an actor hierarchy and use `Save as Prefab...`.
2. Instantiate the prefab by:
   - Drag & drop into Scene View,
   - `Create > From prefab...` from hierarchy/menu bar.
3. Confirm prefab icon appears in hierarchy for the instance root and its children.
4. Modify instance and run `Apply to Prefab`; verify `.ovprefab` content updates.
5. Modify instance and run `Revert to Prefab`; verify actor hierarchy is replaced correctly.
6. Rename/move/delete prefab from Asset Browser; verify references are retargeted in:
   - `.ovscene` files,
   - `.ovprefab` files,
   - current scene actor prefab source references.
7. Sanity-check existing retarget flows (scripts/models/materials/sounds) still work with prefab files included.

Key files:
- `Sources/OvEditor/include/OvEditor/Utils/PrefabOperations.h`
- `Sources/OvEditor/src/OvEditor/Utils/PrefabOperations.cpp`
- `Sources/OvEditor/include/OvEditor/Core/EditorActions.h`
- `Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp`
- `Sources/OvCore/include/OvCore/ECS/Actor.h`
- `Sources/OvCore/src/OvCore/ECS/Actor.cpp`
- `Sources/OvEditor/src/OvEditor/Panels/Hierarchy.cpp`
- `Sources/OvEditor/src/OvEditor/Utils/ActorCreationMenu.cpp`
- `Sources/OvEditor/include/OvEditor/Panels/SceneView.h`
- `Sources/OvEditor/src/OvEditor/Panels/SceneView.cpp`
- `Sources/OvTools/include/OvTools/Utils/PathParser.h`
- `Sources/OvTools/src/OvTools/Utils/PathParser.cpp`
- `Sources/OvCore/src/OvCore/Helpers/GUIHelpers.cpp`
- `Sources/OvEditor/src/OvEditor/Core/EditorResources.cpp`
- `Sources/OvEditor/src/OvEditor/Core/Editor.cpp`
- `Resources/Editor/Textures/Prefab.png`

## AI Usage Disclosure
Generated new code / Refactored code

## Checklist


Sources/OvEditor/src/OvEditor/Core/Editor.cpp
```diff
@@ -131,7 +127,7 @@ void OvEditor::Core::Editor::SetupUI()
 			{
 				EDITOR_EXEC(LoadSceneFromDisk(path));
 			}
-			else if (fileType == EFileType::SCRIPT || fileType == EFileType::SHADER || fileType == EFileType::SHADER_PART)
+			else if (fileType == EFileType::SCRIPT || fileType == EFileType::SHADER || fileType == EFileType::SHADER_PART || fileType == EFileType::PREFAB)
```
adriengivry: Prefabs shouldn't be added here. This are files that are meant to be edited in a code editor, prefabs aren't, so they'll automatically fall back to the next case:
```cpp
// PREFAB, SOUND, FONT, UNKNOWN → open with OS default
OvTools::Utils::SystemCalls::OpenFile(EDITOR_EXEC(GetRealPath(path)));
```

Sources/OvEditor/src/OvEditor/Panels/Hierarchy.cpp
```diff
@@ -73,6 +98,55 @@ class ActorContextualMenu : public OvUI::Plugins::ContextualMenu
...
 			};
+			
+			auto& saveAsPrefabButton = CreateWidget<OvUI::Widgets::Menu::MenuItem>("Save as Prefab...");
+			saveAsPrefabButton.ClickedEvent += [this]
+			{
+				OvWindowing::Dialogs::SaveFileDialog dialog("Save Prefab");
+				const auto initialPath = EDITOR_CONTEXT(projectAssetsPath) / m_target->GetName();
+				dialog.SetInitialDirectory(initialPath.string());
+				dialog.DefineExtension("Overload Prefab", ".ovprefab");
+				dialog.Show();
```
adriengivry, line 135: When "Save as prefab" is used, the name of the actor should be used as the default filename. I recently added `FileDialog::SetInitialFilename()`, so you'll need to rebase to get it on this branch.

adriengivry (changes requested):
Looks really good!

Some feedback I have:
- In Unity, prefabs generally _kind-of_ ignore the transformations on the topmost actor. With your feature, if I make a prefab from an actor, then duplicate this actor many time and place them in a level, reverting them to the original prefab will also move them back to the center of the level. We probably want the "Revert to prefab" to never revert the topmost actor's transform data. The only case the prefab's topmost actor's transform data would be used is when creating a new instance.
- It would be **AWESOME** if we could drag-and-drop any actor from the hierarchy to any folder in the asset browser to turn it into a prefab (using the actor's name as the prefab's name).
- When instancing a prefab, the prefab name should be used for the topmost actor, instead of the actor's name.

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -55,6 +55,73 @@ namespace
...
+
+		const std::string prefabSource = resolvedRoot->GetPrefabSource();
+
+		// Keep compatibility with legacy scenes where every actor of an instance inherited the same source.
+		while (resolvedRoot->HasParent() &&
+			resolvedRoot->GetParent()->HasPrefabSource() &&
+			resolvedRoot->GetParent()->GetPrefabSource() == prefabSource)
+		{
+			resolvedRoot = resolvedRoot->GetParent();
+		}
```
adriengivry: Legacy??? This was never merged.

premake5.lua
```diff
@@ -34,6 +34,7 @@ workspace "Overload"
 	-- Set toolset based on operating system
 	filter {"system:windows"}
 		toolset("msc")
+		buildoptions { "/FS" } -- Prevent C1041 PDB write conflicts with parallel compilation
```
Gopmyc, line 37: I added this build option because it shouldn't cause any problems and could prove useful. I don't think this change requires a separate pull request, even though it's outside the scope of this pull request.

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: It's true 😅

Sources/OvEditor/src/OvEditor/Panels/Hierarchy.cpp
Gopmyc: It's done !

Sources/OvEditor/src/OvEditor/Core/Editor.cpp
Gopmyc: That's taken into account

Sources/OvCore/include/OvCore/SceneSystem/PrefabOperations.h
```diff
@@ -0,0 +1,48 @@
...
+{
+	class Actor;
+}
+
+namespace OvEditor::Utils
+{
+	/**
+	* Utility functions to serialize and instantiate prefab files.
+	*/
+	class PrefabOperations
```
adriengivry, line 22: I feel like this shouldn't live in `OvEditor`, but maybe in `OvCore` instead.
At some point in the future we will want to allow Lua scripts to reference prefabs, and instantiate them, so this functionality will be needed in `OvGame` as well.

Sources/OvEditor/include/OvEditor/Core/EditorActions.h
```diff
@@ -225,6 +225,34 @@ namespace OvEditor::Core
...
+		*/
+		bool ApplyActorToPrefab(OvCore::ECS::Actor& p_actor);
+
+		/**
+		* Revert an actor hierarchy from its prefab source.
+		* The actor instance is replaced by a fresh prefab instantiation.
+		* Returns the new root actor on success.
+		* @param p_actor
+		*/
+		OvCore::ECS::Actor* RevertActorToPrefab(OvCore::ECS::Actor& p_actor);
```
adriengivry: Since this completely replaces the actor, it ends up losing its GUID, as well as all its children GUIDs, breaking any reference to it or its children in any script. Ideally we wouldn't need to delete and recreate the actor completely, and would "only" cleanup the actor: reverting properties, deleting any actor that was added, adding back any actor that was removed, etc...
So this function's signature could look like:
```cpp
bool RevertActorToPrefab(OvCore::ECS::Actor& p_actor);
```

adriengivry (changes requested):
Works very well, I tested with a bunch of Cornell boxes!

Image

<br>

A few issues I found:
- _See comments I added in the code_
- It is hard to know when looking at the hierarchy which source prefab an actor might come from. We could add a field to the actor's inspector, below "GUID" that would show an asset field of type prefab, similar to the "Target" field in asset properties, but that would be fully disabled.
- Reverting an actor to its source prefab overrides its name. In the case where you name your objects (e.g. Pillar 1, Pillar 2, Pillar 3, it will always go back to "Pillar" after reverting)
- Reverting an actor to its source prefab also move it in the hierarchy:

| Before Revert | After Revert |
| - | - |
| Image | Image |

adriengivry (commented):
A regression I also found is using the "Create" menu to add an actor as a child of another actor end up attaching the new actor to the same parent as the actor the contextual menu has been opened from.

Image

| Before | After |
| - | - |
| Image | Image |

Sources/OvCore/include/OvCore/SceneSystem/PrefabOperations.h
Gopmyc: Done !
PrefabOperations has been moved from `OvEditor` to `OvCore` for shared runtime/editor usage

Sources/OvEditor/include/OvEditor/Core/EditorActions.h
Gopmyc: Done !
Revert no longer replaces actors; it now reverts in place and preserves GUID identity

Gopmyc:
> A regression I also found is using the "Create" menu to add an actor as a child of another actor end up attaching the new actor to the same parent as the actor the contextual menu has been opened from.
> 
> Image
> Before	After
> Image	Image

 Good catch !
Fixed the Create menu parenting regression by resolving the contextual parent at click time

Gopmyc:
> Works very well, I tested with a bunch of Cornell boxes!
> 
> Image
> A few issues I found:
> 
> * _See comments I added in the code_
> * It is hard to know when looking at the hierarchy which source prefab an actor might come from. We could add a field to the actor's inspector, below "GUID" that would show an asset field of type prefab, similar to the "Target" field in asset properties, but that would be fully disabled.
> * Reverting an actor to its source prefab overrides its name. In the case where you name your objects (e.g. Pillar 1, Pillar 2, Pillar 3, it will always go back to "Pillar" after reverting)
> * Reverting an actor to its source prefab also move it in the hierarchy:
> 
> Before Revert	After Revert
> Image	Image

Thanks !
Added a disabled Prefab Source field below GUID in Inspector, and revert now preserves actor name and no longer changes hierarchy placement

adriengivry (changes requested):
> [!NOTE]
> Didn't look too hard in the code, since there are a few major issues that prevented further testing, and the code is likely gonna change.

## Incorrect contextual actions
After creating a scene, adding a prefab, and reloading the scene, the contextual menu doesn't show anything else that "Save as prefab..." anymore, even after creating one.

Image

## [Regression] Broken contextual actions on loaded scenes
Whenever a scene is loaded (not new scene), contextual actions on any actor will be broken. e.g. cannot delete, duplicate, save as prefab, "Create..." menu will attach new actor to root instead...

## Wrong GUID re-assignation

1. Create a prefab with actor as children
Image

2. Drag-and-drop first children onto parent
Image

Image

3. Notice A GUID is still correct
Image

4. Revert to prefab
5. Notice A GUID is now different, and C now has A's GUID

Image

Gopmyc:
> Note
> 
> Didn't look too hard in the code, since there are a few major issues that prevented further testing, and the code is likely gonna change.
> 
> ## Incorrect contextual actions
> After creating a scene, adding a prefab, and reloading the scene, the contextual menu doesn't show anything else that "Save as prefab..." anymore, even after creating one.
> 
> Image
> ## [Regression] Broken contextual actions on loaded scenes
> Whenever a scene is loaded (not new scene), contextual actions on any actor will be broken. e.g. cannot delete, duplicate, save as prefab, "Create..." menu will attach new actor to root instead...
> 
> ## Wrong GUID re-assignation
> 1. Create a prefab with actor as children
> 
> Image
> 2. Drag-and-drop first children onto parent
> 
> Image Image
> 3. Notice A GUID is still correct
> 
> Image
> 4. Revert to prefab
> 5. Notice A GUID is now different, and C now has A's GUID
> 
> Image

Thanks for the detailed feedback, this is now addressed.

I fixed the blocking issues in `d05bfd1`:
- Contextual actions no longer break/disappear after scene reload or on loaded scenes: the contextual menu now resolves the target from a live actor pointer validated against the current scene (instead of relying on a stale captured GUID).
- The `Create...` submenu now uses the same live target resolution, so child creation is attached to the correct actor.
- Revert GUID reassignment is fixed: child matching during revert now uses name-based matching first (with positional fallback), preventing GUID swaps when sibling order changed.

Related previous fixes are still in place (`6e47f8d`, `fb1ee23`).

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -55,6 +58,301 @@ namespace
...
+	}
+
+	struct ChildMatchPlan
+	{
+		std::vector<std::pair<size_t, size_t>> matchedChildren;
+		std::vector<size_t> unmatchedTargetChildren;
+		std::vector<size_t> unmatchedTemplateChildren;
+	};
+
+	ChildMatchPlan BuildChildMatchPlan(OvCore::ECS::Actor& p_targetActor, OvCore::ECS::Actor& p_templateActor)
```
adriengivry: The whole matching strategy is still very fragile. We should never rely on actor names, or position in the hierarchy.

Here is an example case that fail:

Initial setup:
```
Stack (GUID 233...)
- A (GUID D72...)
- B (GUID 20B...)
- C (GUID C88...)
```

Moved A below C. And renamed all 3 to Foo:
```
Stack (GUID 233...)
- Foo (previously B) (GUID 20B...)
- Foo (previously C) (GUID C88...)
- Foo (previously A) (GUID D72...)
```

Revert to prefab:
```
Stack (GUID 233...)
- A (GUID 20B...)
- B (GUID C88...)
- C (GUID D72...)
```

As you can see, it doesn't match the initial setup anymore, A is now B, B is now C, and C is now A.

Instead of relying on Actor names and GUIDs, we should add new metadata to actors, alongside `m_prefabSource`, we should have `m_prefabNodeGUID`. Each actor inside of a prefab would have a unique GUID, that would be carried over when instanced. This way, we can easily track back which Actor in a scene correspond to which actor in a prefab.

Sources/OvEditor/include/OvEditor/Utils/ActorCreationMenu.h
```diff
@@ -35,9 +38,9 @@ namespace OvEditor::Utils
         * Generates an actor creation menu under the given MenuList item.
         * Also handles custom additionnal OnClick callback
         * @param p_menuList
-        * @param p_parent
+        * @param p_parentProvider
         * @param p_onItemClicked
         */
-        static void GenerateActorCreationMenu(OvUI::Widgets::Menu::MenuList& p_menuList, OvCore::ECS::Actor* p_parent = nullptr, std::optional<std::function<void()>> p_onItemClicked = {});
+        static void GenerateActorCreationMenu(OvUI::Widgets::Menu::MenuList& p_menuList, ActorParentProvider p_parentProvider = {}, std::optional<std::function<void()>> p_onItemClicked = {});
```
adriengivry: Not sure I understand the point of this `p_parentProvider`. It worked well before without it, with only a pointer to the parent. Why do we suddenly need to patch this?

Sources/OvEditor/src/OvEditor/Panels/Hierarchy.cpp
```diff
@@ -32,52 +36,155 @@
...
-		if (m_target)
+		if (m_targetActor)
 		{
 			auto& focusButton = CreateWidget<OvUI::Widgets::Menu::MenuItem>("Focus");
 			focusButton.ClickedEvent += [this]
 			{
-				EDITOR_EXEC(MoveToTarget(*m_target));
+				if (auto* target = GetTargetActor())
+				{
+					EDITOR_EXEC(MoveToTarget(*target));
```
adriengivry, line 91: Why do we need that? It worked fine before.
This look like a workaround on a problem that hasn't been properly root-caused.

Same for all the actions below.

Sources/OvEditor/src/OvEditor/Panels/Hierarchy.cpp
```diff
@@ -113,18 +226,69 @@ class ActorContextualMenu : public OvUI::Plugins::ContextualMenu
...
+		{
+			return nullptr;
+		}
+
+		const auto& actors = currentScene->GetActors();
+		const bool actorIsStillInScene =
+			std::find(actors.begin(), actors.end(), m_targetActor) != actors.end();
+
+		return actorIsStillInScene ? m_targetActor : nullptr;
+	}
```
adriengivry, line 277: Why can't we just use `OvCore::ECS::Actor* m_target;` like before? I don't understand the problem this is trying to solve.

Sources/OvEditor/src/OvEditor/Utils/ActorCreationMenu.cpp
```diff
@@ -120,69 +116,153 @@ namespace
...
+		return [p_parentProvider = std::move(p_parentProvider), p_onItemClicked]()
+		{
+			CreateCharacter(ResolveParent(p_parentProvider));
+
+			if (p_onItemClicked.has_value())
+			{
+				p_onItemClicked.value()();
+			}
+		};
 	}
```
adriengivry: We shouldn't need to change these, they are working perfectly fine. Same as above, looks like a patch on an issue that hasn't been properly root-caused.

Sources/OvEditor/src/OvEditor/Utils/ActorCreationMenu.cpp
```diff
@@ -120,69 +116,153 @@ namespace
...
+	audio.CreateWidget<MenuItem>("Audio Source").ClickedEvent += ActorWithComponentCreationHandler<CAudioSource>(p_parentProvider, p_onItemClicked);
+	audio.CreateWidget<MenuItem>("Audio Listener").ClickedEvent += ActorWithComponentCreationHandler<CAudioListener>(p_parentProvider, p_onItemClicked);
+	others.CreateWidget<MenuItem>("Camera").ClickedEvent += ActorWithComponentCreationHandler<CCamera>(p_parentProvider, p_onItemClicked);
+	others.CreateWidget<MenuItem>("Post Process Stack").ClickedEvent += ActorWithComponentCreationHandler<CPostProcessStack>(p_parentProvider, p_onItemClicked);
+	others.CreateWidget<MenuItem>("Reflection Probe").ClickedEvent += ActorWithComponentCreationHandler<CReflectionProbe>(p_parentProvider, p_onItemClicked);
+	others.CreateWidget<MenuItem>("Skysphere").ClickedEvent += CreateSkysphereHandler(p_parentProvider, p_onItemClicked);
+	others.CreateWidget<MenuItem>("Atmosphere").ClickedEvent += CreateAtmosphereHandler(p_parentProvider, p_onItemClicked);
+	others.CreateWidget<MenuItem>("Character").ClickedEvent += CreateCharacterHandler(p_parentProvider, p_onItemClicked);
```
adriengivry: Shouldn't be needed, see comment above.

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
```diff
@@ -882,6 +1181,139 @@ void OvEditor::Core::EditorActions::DuplicateActor(OvCore::ECS::Actor & p_toDupl
...
+
+	if (instantiatedRoot)
+	{
+		SetRootPrefabSourceAndNormalizeChildren(*instantiatedRoot, prefabSourcePath);
+
+		const std::string prefabName = realPath.stem().string();
+		if (!prefabName.empty())
+			instantiatedRoot->SetName(prefabName);
+
+		OVLOG_INFO("Prefab instantiated: " + realPath.string());
```
adriengivry: Not needed. We recently removed the "Actor created" log, this one should follow the same treatment.

adriengivry (changes requested):
Functionality-wise it seems to be nearly feature-complete. There are a few bugs and issues I found that are listed below.

Otherwise great work!

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: Implemented. Revert matching now uses actor prefab_node_guid metadata (persisted in prefab actors and propagated on instantiation), not actor names or hierarchy position. I also fixed ActorRef remapping by mapping both prefab-node GUIDs and serialized prefab GUIDs to current instance GUIDs, which covers the Apply -> Revert case correctly

Sources/OvEditor/include/OvEditor/Utils/ActorCreationMenu.h
Gopmyc: Good point. p_parentProvider has been removed and the API is back to a plain OvCore::ECS::Actor* p_parent, as before

Sources/OvEditor/src/OvEditor/Panels/Hierarchy.cpp
Gopmyc: Root cause was stale actor pointers when a context menu stays open across scene mutations (delete/reload). Actions now resolve the target via GetTargetActor() and execute only if the actor is still in the current scene

Sources/OvEditor/src/OvEditor/Panels/Hierarchy.cpp
Gopmyc: We are using OvCore::ECS::Actor* m_targetActor again. The only addition is a lightweight validity check before executing actions

Sources/OvEditor/src/OvEditor/Utils/ActorCreationMenu.cpp
Gopmyc: I removed the broader workaround approach and kept the menu contract simple. The remaining change is a minimal click-time ResolveAliveParent guard so creation safely falls back if the parent was removed

Sources/OvEditor/src/OvEditor/Utils/ActorCreationMenu.cpp
Gopmyc: Applied the same minimal parent-validity guard consistently to all creation entries (including Create Empty) to avoid inconsistent behavior

Sources/OvEditor/src/OvEditor/Core/EditorActions.cpp
Gopmyc: Done. I removed the non-essential prefab apply/revert info logs to align with the recent log cleanup policy

adriengivry (approved):
LGTM!

adriengivry:
Seems like there are a few conflicts following the merging of your other PRs.

Gopmyc:
> Seems like there are a few conflicts following the merging of your other PRs.

It's sorted !

PR #778, opened 2026-04-24
Add `Actor.GetComponent(type)` and align `Actor.GetBehaviour(type)` API

### LINKED ISSUE #777: Add `GetComponent(type)` and `GetBehaviour(type)` to the Actor Lua bindings
The current `Actor` Lua API exposes many hardcoded getters (`GetCamera`, `GetLight`, etc.) and `GetBehaviour(name)`, but there is no generic lookup by type. This makes scripts more verbose, less scalable, and harder to write in a data-driven way when the component/behaviour type is only known at runtime.

## Expected solution
Add two generic `Actor` bindings in Lua:
- `GetComponent(type)`
- `GetBehaviour(type)`

Both should resolve from a type identifier and return the matching instance, or `nil` when not found. Existing explicit getters should remain supported for backward compatibility.

[Related to #689]

Gopmyc:
## Description
This PR adds a generic `Actor:GetComponent(type)` binding to the Lua API and updates the Lua Actor stubs accordingly.

Changes included:
- Added `Actor:GetComponent(type)` in Lua bindings (`LuaActorBindings.cpp`).
- Implemented a type-to-resolver registry for cleaner and maintainable component lookup.
- Kept behavior strict: unknown component type returns `nil`.
- Aligned `GetBehaviour` parameter naming to `type` in bindings and Lua stubs for API consistency.
- Updated `Resources/Engine/Lua/Scene/Actor.lua` to expose:
  - `Actor:GetComponent(type)`
  - `Actor:GetBehaviour(type)`

## Related Issue(s)
Fixes #777

## Review Guidance
Please focus review on:
- Type-based component lookup behavior in `Actor:GetComponent(type)`.
- Returned value semantics (`nil` when type is unknown or missing).
- API consistency between C++ bindings and Lua stubs.

## Screenshots/GIFs
N/A

## AI Usage Disclosure
Generated new code / Documentation (I did the refactoring)

## Checklist

### DIFF
--- a/Resources/Engine/Lua/Scene/Actor.lua
+++ b/Resources/Engine/Lua/Scene/Actor.lua
@@ -136,10 +136,15 @@ function Actor:GetPostProcessStack() end
 ---@return ReflectionProbe|nil
 function Actor:GetReflectionProbe() end
 
+--- Returns the component of the given type attached to this actor (If any)
+---@param type string
+---@return Component|nil
+function Actor:GetComponent(type) end
+
 --- Returns the Behaviour of the given type attached to this actor (If any)
----@param name string
+---@param type string
 ---@return table|nil
-function Actor:GetBehaviour(name) end
+function Actor:GetBehaviour(type) end
 
 --- Adds a Transform component to the actor and returns it
 ---@return Transform
--- a/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaActorBindings.cpp
+++ b/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaActorBindings.cpp
@@ -4,7 +4,10 @@
 * @licence: MIT
 */
 
+#include <algorithm>
+#include <array>
 #include <filesystem>
+#include <string_view>
 
 #include <sol/sol.hpp>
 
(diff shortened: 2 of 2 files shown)


adriengivry:
Not exactly what I had in mind.

By `type` I didn't mean a string, but a type (i.e. table name).

For example, in `Transform.lua`:
```lua
---@meta

--- Represents the 3D transformations applied to an actor
---@class Transform : Component
Transform = {}
```
That should allow us to do something like that in a script:
```lua
function MyScript:OnStart()
    self.owner:GetComponent(Transform)
end
```
Since `Transform` is a valid auto-completable type (thanks to the LSP), it would make it very convenient and safe (prevent typos etc...).

Not sure if it's even possible, but we should definitely investigate.

Gopmyc:
> Not exactly what I had in mind.
> 
> By `type` I didn't mean a string, but a type (i.e. table name).
> 
> For example, in `Transform.lua`:
> 
> ```lua
> ---@meta
> 
> --- Represents the 3D transformations applied to an actor
> ---@class Transform : Component
> Transform = {}
> ```
> 
> That should allow us to do something like that in a script:
> 
> ```lua
> function MyScript:OnStart()
>     self.owner:GetComponent(Transform)
> end
> ```
> 
> Since `Transform` is a valid auto-completable type (thanks to the LSP), it would make it very convenient and safe (prevent typos etc...).
> 
> Not sure if it's even possible, but we should definitely investigate.

That would mean polluting Lua's global table with lots of elements like `Transform`, is that really a good idea ?

adriengivry:
@Gopmyc wdym pollute the global table? These user types are already defined in the global table: [LuaComponentBindings.cpp](https://github.com/Overload-Technologies/Overload/blob/main/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaComponentsBindings.cpp) and captured by the LSP thank to the LuaLS definitions

Gopmyc:
> @Gopmyc wdym pollute the global table? These user types are already defined in the global table: [LuaComponentBindings.cpp](https://github.com/Overload-Technologies/Overload/blob/main/Sources/OvCore/src/OvCore/Scripting/Lua/Bindings/LuaComponentsBindings.cpp) and captured by the LSP thank to the LuaLS definitions

Fair point. You’re right about the current state: these types are already exposed in the global table and LuaLS relies on that.  
I still think there’s an architecture discussion worth having for later: the more we expose in globals, the more we risk name collisions, cross-script side effects, and harder long-term maintenance.  
For now I’ll close this PR, and I’ll come back with a cleaner/aligned proposal around that

PR #812, opened 2026-05-18
Fix editor shortcuts during play mode

### LINKED ISSUE #804: Save shortcut still works while the scene is playing
## Description
When a scene is running inside the game engine, the keyboard shortcut used to save the scene is still active.

This behavior can interfere with gameplay inputs, especially if the same key combination or one of its keys is used by the game. Editor shortcuts should be disabled or ignored while the scene is playing, so the game can fully handle keyboard input without conflict.

## To Reproduce
1. Open a scene in Overload.
2. Start playing the scene in the game engine.
3. Press the keyboard shortcut used to save the scene.
4. Notice that the scene save action is triggered even though the scene is currently playing.

## Expected behavior
The save scene shortcut should not work while the scene is playing.

Editor shortcuts should be disabled during play mode to avoid interfering with global game controls and player input.

## Screenshots
N/A

Gopmyc:
## Description
Prevents editor keyboard shortcuts from running outside edit mode, so Ctrl+S and related scene shortcuts no longer affect the runtime scene while playing.

## Related Issue(s)
Fixes #804

## Review Guidance
Check the edit-mode guards in `Editor::Update` and `MenuBar::HandleShortcuts`.

## Screenshots/GIFs
N/A

## AI Usage Disclosure
N/A

## Checklist

### DIFF
--- a/Sources/OvEditor/src/OvEditor/Core/Editor.cpp
+++ b/Sources/OvEditor/src/OvEditor/Core/Editor.cpp
@@ -229,7 +229,11 @@ void OvEditor::Core::Editor::Update(float p_deltaTime)
 	const bool mouseEnabled = m_context.window->GetCursorMode() != OvWindowing::Cursor::ECursorMode::DISABLED;
 	m_context.uiManager->EnableMouse(mouseEnabled);
 
-	HandleGlobalShortcuts();
+	if (m_editorActions.GetCurrentEditorMode() == EditorActions::EEditorMode::EDIT)
+	{
+		HandleGlobalShortcuts();
+	}
+
 	UpdateCurrentEditorMode(p_deltaTime);
 	RenderViews(p_deltaTime);
 	UpdateEditorPanels(p_deltaTime);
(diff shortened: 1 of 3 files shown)


adriengivry (changes requested):
Most of these shortcuts should remain available even in play mode.
We should always be able to:
- Change scene (CTRL + N)
- Delete an object
- etc..

The only actions to disable are `EditorActions::SaveSceneChanges()` and `EditorActions::SaveAs()`.

Adding this to both these functions should suffice:
```cpp
if (m_editorMode != EEditorMode::EDIT)
{
    return;
}
```

Gopmyc:
> Most of these shortcuts should remain available even in play mode. We should always be able to:
> 
> * Change scene (CTRL + N)
> * Delete an object
> * etc..
> 
> The only actions to disable are `EditorActions::SaveSceneChanges()` and `EditorActions::SaveAs()`.
> 
> Adding this to both these functions should suffice:
> 
> ```c++
> if (m_editorMode != EEditorMode::EDIT)
> {
>     return;
> }
> ```

Indeed…
It's fixed !

adriengivry (approved):

PR #818, opened 2026-05-19
Added freetype and updated font to Roboto

### LINKED ISSUE #681: Add freetype fonts with icon support
Fonts don't render super well when very small. This is because we use ImGui's truetype implementation, instead of freetype.

## Expected solution
Update ImGui to use freetype. Ideally we would want to be able to support icons (like font-awesome?) or maybe nerd fonts.

I guess we should build freetype as a dependency.

How to doc: https://github.com/ocornut/imgui/blob/master/misc/freetype/README.md

adriengivry:
## Description

- Added freetype
- Updated ImGui to use freetype instead of truetype
- Changed default font from Ruda-Bold to Roboto

## Related Issue(s)

Fixes #681 

## Review Guidance

Write here.

## Screenshots/GIFs

Screenshot_20260519_153317

_Before (truetype + Ruda-Bold)_

Screenshot_20260519_153325

_After (freetype + Roboto)_

## AI Usage Disclosure

None

## Checklist


Gopmyc:
Wouldn’t it be better to merge the PR into a test branch instead of `main` while we run tests and confirm that the new dependency, or its integration, doesn’t introduce any new errors ?

adriengivry:
> Wouldn’t it be better to merge the PR into a test branch instead of `main` while we run tests and confirm that the new dependency, or its integration, doesn’t introduce any new errors ?

It's fairly safe. Easy to revert if anything goes wrong, and having it on main gives us an opportunity to test it more thoroughly. Most people serious about ImGui actually use freetype instead of truetype, so it's not like we're doing anything weird or likely to break here.

Gopmyc:
> > Wouldn’t it be better to merge the PR into a test branch instead of `main` while we run tests and confirm that the new dependency, or its integration, doesn’t introduce any new errors ?
> 
> It's fairly safe. Easy to revert if anything goes wrong, and having it on main gives us an opportunity to test it more thoroughly. Most people serious about ImGui actually use freetype instead of truetype, so it's not like we're doing anything weird or likely to break here.

Okay, I trust you then, I'm just suggesting it in case

PR #825, opened 2026-06-17
Make glGet state queries buffer-safe

### LINKED ISSUE #644: Unsafe `GetInt`, `GetBool`, `GetFloat`, and `GetDouble` leading to potential buffer overruns
## Description
The current implementation of our `GetInt`, `GetBool`, `GetFloat`, and `GetDouble` functions is unsafe.

This has been partially addressed by https://github.com/Overload-Technologies/Overload/pull/640 which proposed a workaround for the `GetInt` which is currently being used, and therefore required immediate attention.

This issue has been addressed on [baregl](https://github.com/adriengivry/baregl) (modern OpenGL wrapper) by this PR: https://github.com/adriengivry/baregl/pull/10

## Expected Solution
We need to provide a way to specify the buffer size, while providing ways to retrieve single values, arrays, and add support for indexed get parameters.

SixSleepyAsh:
## Summary
- Replaces raw `glGet*` calls with buffer-safe variants that avoid undefined behavior on mismatched return types
- Ensures all state queries write into correctly-sized buffers

Fixes #644

### DIFF
--- a/Sources/OvRendering/src/OvRendering/HAL/OpenGL/GLBackend.cpp
+++ b/Sources/OvRendering/src/OvRendering/HAL/OpenGL/GLBackend.cpp
@@ -4,7 +4,11 @@
 * @licence: MIT
 */
 
+#include <array>
+#include <concepts>
+#include <cstdint>
 #include <format>
+#include <span>
 
 #include <glad.h>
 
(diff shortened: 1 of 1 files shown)


adriengivry:
This will already be fixed by #824

Overload isn't accepting any changes to its HAL at the moment.

Thanks for your contribution!

PR #834, opened 2026-08-30
Support Animation Layers for Skinned Meshes

### LINKED ISSUE #833: Animation Layers - Support for simultaneous skeletal animations with blending

Currently, `CSkinnedMeshRenderer` can only play one animation at a time. This limitation prevents developers from:

1. **Layering independent animations**: Cannot play multiple animation tracks simultaneously (e.g., base locomotion + upper body action, or idle + gesture)
2. **Reusing animation clips across contexts**: Animation combinations must be pre-baked as separate clips instead of dynamically composed
3. **Smooth animation blending**: No built-in support for weighted blending between multiple poses

This results in animation clip duplication, increased asset management complexity, and reduced runtime flexibility.

## Expected solution

Implement an **Animation Layers** system in `CSkinnedMeshRenderer`:

**Features:**
- Multiple simultaneous animation layers per renderer
- Each layer operates independently with its own playback state:
  - `animationIndex` / `animationName`
  - `currentTimeTicks` (per-layer animation time)
  - `playbackSpeed` (per-layer speed control)
  - `isPlaying` / `isLooping` (per-layer playback mode)
  - `weight` (0.0–1.0, contribution to final pose blending)
- Blend skeletal poses at the bone level using:
  - Linear interpolation (Lerp) for position and scale channels
  - Spherical linear interpolation (Slerp) for rotation channels
- Automatic weight normalization (sum to 1.0)

Gopmyc:
## Description
`CSkinnedMeshRenderer` played a single clip. It now drives up to 4 layers, each owning its
animation source model, animation, time, speed, loop mode and weight. Layer-scoped methods take a
trailing layer index defaulting to the base layer, which always exists

Pose evaluation iterates skeleton nodes instead of animation tracks, resolving each layer's track
per node. Only layers holding a track for a node contribute to it, their weights normalized against
each other, lerp for position and scale, slerp for rotation. Nodes no layer animates keep their
bind pose

## Related Issue(s)
Fixes #833

## Review Guidance
- A single layer at any weight above zero reproduces the previous output; zero active layers gives
  the bind pose.
- Weights normalize per node, so a layer that alone drives a bone stays at full strength until its
  weight reaches zero, where the bone falls back to bind pose. Documented on `SetLayerWeight`.
- Going node-major also makes duplicate tracks pointing at the same node harmless, where
  accumulating weights would have counted them twice. `AssimpParser` can emit those.
- **Scene format break:** per-layer settings moved under a `<layers>` element. Existing scenes lose
  `playing`, `looping`, `playback_speed`, `time_ticks`, `animation` and `animation_source`, and
  reset to one default layer. No migration path.
- `Columns` only drew `Separator` full width, through a hardcoded `dynamic_cast`. That is now an
  opt-in `AWidget::fullWidth` flag, required for the collapsible layer groups. Default false, no
  existing widget affected.
- No bone masks: a layer only affects nodes its clip has tracks for. A rotation-only track still
  contributes bind translation, matching the previous single-clip behaviour.

## Screenshots/GIFs
N/A

## AI Usage Disclosure
Documentation + Refactored code

## Checklist


Resources/Engine/Lua/Components/SkinnedMeshRenderer.lua
```diff
@@ -1,41 +1,83 @@
...
---- Returns whether playback is active
+--- Removes an animation layer, shifting the following layers down by one.
+--- Fails when the index is invalid or when only one layer remains
+---@param layer integer
 ---@return boolean
-function SkinnedMeshRenderer:IsPlaying() end
+function SkinnedMeshRenderer:RemoveLayer(layer) end
+
+--- Starts/resumes animation playback on a layer
+---@param layer? integer
```
Gopmyc, line 29: '?' marks the param as optional in LuaLS. First time it shows up in the repo, can switch to 'integer|nil' if you'd rather ?

Sources/OvCore/include/OvCore/ECS/Components/CSkinnedMeshRenderer.h
```diff
@@ -22,8 +23,14 @@ namespace OvRendering::Resources { class Model; }
 
 namespace OvCore::ECS::Components
 {
+	constexpr uint32_t kMaxAnimationLayers = 4;
```
Gopmyc: Arbitrary limit, raise if needed ?

Sources/OvUI/include/OvUI/Widgets/Layout/Columns.h
```diff
@@ -122,7 +122,7 @@ namespace OvUI::Widgets::Layout
 
 			for (auto it = m_widgets.begin(); it != m_widgets.end(); ++it)
 			{
-				if (dynamic_cast<Visual::Separator*>(it->first))
+				if (it->first->fullWidth || dynamic_cast<Visual::Separator*>(it->first))
```
Gopmyc, line 125: Needed for the collapsible layer groups : The inspector pins column 0 at 200px, so a
GroupCollapsable dropped in a cell gets crushed. Only 'Separator' could break out of the table
before, via this hardcoded cast, the flag just generalizes it. Defaults to false, nothing else
changes

Gopmyc (commented):
Just a few small changes to clarify things that I didn't find particularly explicit in the code itself

Resources/Engine/Lua/Components/SkinnedMeshRenderer.lua
adriengivry: I'm fine to start using `type?` instead of `type|nil`, good catch!
We'll update other `type|nil` to `type?` as we go

Sources/OvCore/include/OvCore/ECS/Components/CSkinnedMeshRenderer.h
adriengivry: Any idea what it looks like in other engines/animation systems? Is 4 standard?

adriengivry: Do we need a limit? Maybe an `std::vector` of layers instead of an `std::array` would make more sense, if the performance hit isn't too high

adriengivry (commented):
I like the feature, I think it's a great addition.
The UI is a little confusing, "Layer 0", "Layer 1", etc... look like other components.

Image

We probably want to keep collapsible groups for components, and find another way to present array elements.

Resources/Engine/Lua/Components/SkinnedMeshRenderer.lua
Gopmyc: I think using AI to do that could be a huge time saver

Sources/OvCore/include/OvCore/ECS/Components/CSkinnedMeshRenderer.h
Gopmyc: Good call, switched to a `std::vector`
I'd assumed 4 was plenty and it made the implementation easier, but it's not really my call to
make, better to let people lay out their animations how they want

Resources/Engine/Lua/Components/SkinnedMeshRenderer.lua
adriengivry: Yeah AI is great for refactoring code like that :)
Could be done in another PR

Sources/OvCore/src/OvCore/ECS/Components/CSkinnedMeshRenderer.cpp
```diff
@@ -757,79 +892,209 @@ void OvCore::ECS::Components::CSkinnedMeshRenderer::OnInspector(OvUI::Internal::
...
-	auto& animationChoice = p_root.CreateWidget<OvUI::Widgets::Selection::ComboBox>(currentAnimIndex);
-
-	auto& animDispatcher = animationChoice.AddPlugin<OvUI::Plugins::DataDispatcher<int>>();
-	animDispatcher.RegisterGatherer([this, &animationChoice]
+	auto& modelDiagnostic = p_root.CreateWidget<OvUI::Widgets::Texts::Text>();
+	modelDiagnostic.AddPlugin<OvUI::Plugins::DataDispatcher<std::string>>().RegisterGatherer([this]
 	{
-		animationChoice.choices.clear();
-		animationChoice.choices.emplace(-1, "<None>");
+		return HasCompatibleModel() ? std::string{} : std::string{ "No skinned model assigned" };
```
adriengivry: I like the idea of having a diagnostic message here, but when no issue is found, it leaves a blank spot. 


Image

Either we could add a message like for scripts, or disable the text dynamically.

Image

Gopmyc: Added a `Ready` at the component level like `Behaviour` does, and colored the layer messages. Left
the layers quiet when they're fine though, repeating "Ready" on each one gets noisy fast

Tried disabling the field first, but `ExecutePlugins` runs inside the `if (enabled)` block in
`AWidget::Draw`, so a widget that hides itself from its own gatherer never comes back. Probably
worth fixing on its own

adriengivry: Already much better!
I think the diagnostic message should have 2 columns, same as scripts.

"Ready" in the first column, should have a text message in the second column that says something like: "Compatible skinned model found"

When not compatible, it should say "Error" in read in the first column, and "No skinned model assigned".

image

Gopmyc: Done ! 🙌

adriengivry: Looks great! I guess the same logic should apply here:
image

Either:
- Ready | Compatible animation source found
- Error | Animation source skeleton is not compatible with model
- Warning | Animation source has no animation clips
- Warning | Model has no animation clips

Maybe "Animation Source" could also be renamed in the inspector to "Animation Source Override", so users know it's not mandatory

Gopmyc: Got it

adriengivry (approved):

PR #837, opened 2026-09-05
Allow resizing of windows

tccountus:
## Description
This PR allows resizing of windows. Maybe we could also add a Lua-API to change window settings

## Review Guidance
- Framebuffer is resized
- ~~Widgets/Panels are currently not resized correctly~~

## Screenshots/GIFs
output

## AI Usage Disclosure
None

## Checklist
- [ ] When applicable, I have commented my code, particularly in hard-to-understand areas
- [x] ~~Currently, the debug driver info panel is not resized and moved properly~~

One thing: the project settings gets resetted every time we update them, I think we should change that in the future.


adriengivry (changes requested):
Tested on Arch, works well. This PR should also address the issue with hardware info losing their alignment after resizing. `HardwareInfo.cpp` should be updated to `SetAlignment` in a similar fashion as `FPSCounter.cpp` and `FrameInfo.cpp`

adriengivry (approved):
LGTM!

PR #842, opened 2026-09-09
Prevent native crashes on invalid Lua calls in optimized builds

### LINKED ISSUE #828: Crash in OvEditor calling Scene:FindActorByName from Lua : access violation in sol (null userdata)
## Description

Calling Scene::FindActorByName from Lua causes OvEditor.exe to crash with an access violation in sol (stack_get_unqualified.hpp). Crash occurs when sol dereferences the internal userdata pointer (void* udata = *pudata), reading 0x00000000. The failure happens while resolving the bound Scene usertype when the Lua call is dispatched.

## To Reproduce

1.	Launch OvEditor
2.	In Lua, execute: Debug.Log(tostring(Scene:FindActorByName("...")))
3.	The engine attempts to call the bound Scene member function.
4.	OvEditor.exe crashes with an access violation inside sol at Dependencies/sol/include/sol/stack_get_unqualified.hpp (dereference of userdata pointer).

## Expected behavior

A stable call that returns either a valid Actor userdata or nil/string result; no native crash. If the Lua call is invalid (nil self or wrong call syntax), sol should return a clear error rather than dereferencing a null pointer.

## Screenshots
N/A

## Additional useful details (to include in the GitHub issue):
•	Exception: 0xC0000005 (Access violation reading location 0x00000000)
•	Fault site: solstackunqualified_getter<...>::get_no_lua_nil — line where void* udata = *pudata; is executed.
•	Likely triggers: lua_touserdata(L, index) returned null (light userdata, nil, or wrong stack index); member call invoked without a valid 'self' userdata; mismatch between new_usertype registration and bound member functions; or userdata lifetime corruption.
•	Immediate diagnostic steps: inspect Lua stack at the call (type_of(L,1), lua_touserdata(L,1)); confirm Lua callsite uses ':'; enable SOL_SAFE_USERTYPE to get safer error messages.

Gopmyc:
## Description
sol2 only enables its safety checks on debug builds. Release and Publish compile with `NDEBUG`, so `SOL_SAFE_USERTYPE` is off and a member call with a bad `self` goes down the unchecked branch in `call.hpp`. That branch dereferences a null userdata pointer in `stack_get_unqualified.hpp`, so we get an access violation instead of a Lua error.

`Scene:FindActorByName("...")` is one way to reach it. The global `Scene` is the usertype metatable, not a scene instance (the instance comes from `Scenes.GetCurrentScene()`). Any bound member of the 36 registered usertypes does the same thing when `self` is wrong, and so do nil or mistyped arguments.

I fixed it with a build define rather than in the bindings because sol2 already has the check, we just compile it out. Doing it by hand would mean validating `self` on every bound member of every usertype, then remembering it on each new binding. sol2 points to build defines for this in `sol/config.hpp`, and patching the vendored header would be lost the next time we update the dependency.

One thing to know before merging: Release and Publish now run integer precision checks as well. `Math.RandomInt(1, n / 2)` raises an error instead of silently truncating, since `/` returns a float in Lua 5.3+. Debug builds have always behaved this way.

## Related Issue(s)
Fixes #828

## Review Guidance
- Placed at workspace level to match `SOL_NO_LUA_HPP` from #692. All sol2 translation units currently sit in OvCore, so scoping the define to that project works just as well if you prefer it there.
- Cost is one type check per `self` and per argument on every bound call, in all configurations.
- If the integer strictness is unwanted, `SOL_SAFE_USERTYPE` and `SOL_SAFE_FUNCTION_CALLS` on their own remove the crash without touching number handling.
- This does not cover crashes sol2 has no say in, such as `Scenes.GetCurrentScene()` dereferencing a null scene.

## Screenshots/GIFs
N/A

## AI Usage Disclosure
Debugging

## Checklist

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -19,6 +19,7 @@ workspace "Overload"
 	defines {
 		"LUA_SCRIPTING",
 		"OVERLOAD_VERSION=\"" .. version .. "\"",
+		"SOL_ALL_SAFETIES_ON",
 		"SOL_NO_LUA_HPP"
 	}
 


Gopmyc (commented):
The crash needs fixing, I'm sure of that. Less sure `SOL_ALL_SAFETIES_ON` is the tradeoff you want, given the cost in every config. Would rather have someone else weigh in

PR #851, opened 2026-09-14
Fix crash when creating actors from OnStart

### LINKED ISSUE #844: Crash when creating actors from OnStart
## Description
Creating actors from a script's `OnStart` (same for `OnAwake` and `OnEnable`) can crash the engine when entering play mode.

`Scene::Play` loops over `m_actors` with iterators while calling these callbacks. If a script calls `Scene:CreateActor` there, `m_actors` can be reallocated and the loop keeps reading the old buffer, which ends in an access violation in `Actor::IsActive`.

It only crashes when the actor holding the script isn't the last one in the scene, because the loop stops right after the last actor. `Scene::Update`, `FixedUpdate` and `LateUpdate` don't have this problem since they iterate over a copy of `m_actors`. Doing the same in `Scene::Play` should fix it.

## To Reproduce
1. Create a new scene (File > New Scene)
2. Create a script with the following content:
```lua
local Spawner = {}

function Spawner:OnStart()
    for i = 1, 10 do
        Scenes.GetCurrentScene():CreateActor("Spawned " .. i, "")
    end
end

return Spawner
```
3. Add the script to `Main Camera`, which is the first actor of the scene
4. Press Play
5. See the crash (access violation in `Actor::IsActive`)

## Expected behavior
The actors are created and play mode starts normally, the same way it works when `CreateActor` is called from `OnUpdate`.

## Screenshots
N/A

Element119rd:
## Description
Fixes a crash when entering play mode caused by creating actors from `OnAwake`, `OnEnable` or `OnStart`.

`Scene::Play` iterated over `m_actors` directly while invoking these callbacks. If a script called `Scene::CreateActor` from one of them, `m_actors` could be reallocated mid-loop, invalidating the iterators and causing an access violation in `Actor::IsActive`. It only crashed when the actor holding the script wasn't the last one in the scene, because the loop stopped right after the last actor.

`Scene::Play` now iterates over a copy of `m_actors`, the same way `Update`, `FixedUpdate` and `LateUpdate` already do. The copy is taken once before all three callbacks so actors created during `OnAwake` don't get `OnEnable`/`OnStart` called twice.

While fixing this, the `BeginBatchActorCreation`/`EndBatchActorCreation` `Play`'s callbacks was removed. `CreateActor` already starts actors created during play immediately, so the batch was redundant, and it caused a nested-batch assert when a callback called `Scene::InstantiatePrefab`. `EndBatchActorCreation` is reverted to its previous shape: it swaps the pending actors into a local list and clears `m_batchActorCreation` before running any callback, which removes the nesting problem and makes the `while` loop unnecessary.

## Related Issue(s)
Fixes #844

## Review Guidance
The only functional change is in `Scene::Play` 

## Screenshots/GIFs
None!

## AI Usage Disclosure
None!

## Checklist


Sources/OvCore/src/OvCore/SceneSystem/Scene.cpp
```diff
@@ -200,42 +207,46 @@ void OvCore::SceneSystem::Scene::EndBatchActorCreation(bool p_startCreatedActors
...
+			{
+				if (actor.get().IsActive())
+				{
+					actor.get().OnStart();
+				}
 			}
 		}
 	}
+
+	m_batchActorCreation = false;
```
Gopmyc: Moving `m_batchActorCreation = false` after the loop means the batch is still active while `OnAwake`/`OnEnable`/`OnStart` run. If one of these callbacks calls `Scene:InstantiatePrefab` (e.g. a prefab whose script instantiates another prefab in `OnStart`), `BeginBatchActorCreation` hits the "Cannot begin an actor creation batch while another one is still active" assert. In Release the assert is compiled out, and `m_batchCreatedActors` gets cleared, so actors created earlier in the same callback are never started

This works on `main`, since the batch is swapped out and the flag reset before invoking callbacks. Actors created from these callbacks already go through `CreateActor`'s immediate path, so I don't think the `while` loop is needed. Could we revert `EndBatchActorCreation` to its previous version ?

Sources/OvCore/src/OvCore/SceneSystem/Scene.cpp
```diff
@@ -121,9 +121,16 @@ void OvCore::SceneSystem::Scene::Play()
 	/* Wake up actors to allow them to react to OnEnable, OnDisable and OnDestroy, */
 	std::for_each(m_actors.begin(), m_actors.end(), [](ECS::Actor * p_element) { p_element->SetSleeping(false); });
 
-	std::for_each(m_actors.begin(), m_actors.end(), [](ECS::Actor * p_element) { if (p_element->IsActive()) p_element->OnAwake(); });
-	std::for_each(m_actors.begin(), m_actors.end(), [](ECS::Actor * p_element) { if (p_element->IsActive()) p_element->OnEnable(); });
-	std::for_each(m_actors.begin(), m_actors.end(), [](ECS::Actor * p_element) { if (p_element->IsActive()) p_element->OnStart(); });
+	/* Batch actor creation to handle actors created during initialization callbacks */
+	BeginBatchActorCreation();
```
Gopmyc: With `Play()` running its callbacks inside a batch, calling `Scene:InstantiatePrefab` from any `OnAwake`/`OnEnable`/`OnStart` when entering play mode triggers the same assert

The copy below should be enough to fix #844 on its own : `CreateActor` starts the new actor immediately, and since it isn't part of the copy, it won't receive its callbacks twice. That's also what `Update`, `FixedUpdate` and `LateUpdate` do. Could we drop `BeginBatchActorCreation`/`EndBatchActorCreation` here and keep only the copy ?

Gopmyc (changes requested):
NIT : the PR title ends up as the squashed commit message, and the crash happens when entering play mode rather than during loading. Maybe something closer to the issue title, like "Fix crash when creating actors from OnStart" ?

Sources/OvCore/src/OvCore/SceneSystem/Scene.cpp
Element119rd: Done , dropped the batching from Play() and reverted EndBatchActorCreation so the flag is cleared before callbacks run.

Sources/OvCore/src/OvCore/SceneSystem/Scene.cpp
Element119rd: done. dropped BeginBatchActorCreation/EndBatchActorCreation from Play() and kept only the copy.

adriengivry (changes requested):
Thanks for contributing! 🎉

Please use the standard PR template https://github.com/Overload-Technologies/Overload/blob/main/.github/pull_request_template.md

Sources/OvCore/src/OvCore/SceneSystem/Scene.cpp
Gopmyc: Thanks ! One small thing left : the swap is now inside the if, so m_batchCreatedActors isn't cleared when the batch isn't started (edit mode, or EndBatchActorCreation(false)). It's harmless since nothing reads it before the next BeginBatchActorCreation, but since Play() no longer uses the batch, this function doesn't need to change at all. Could we restore its previous version, so the PR only touches Play()?

Element119rd: restored EndBatchActorCreation to its version on main

Gopmyc: Perfect, you can mark the comment as resolved

Sources/OvCore/src/OvCore/SceneSystem/Scene.cpp
Gopmyc: Nice, you can also mark the comment as resolved

Gopmyc (approved):
LGTM ! ✅

adriengivry (approved):

PR #852, opened 2026-09-14
Expose the size of the rendered area to Lua

### LINKED ISSUE #845: Inputs.GetMousePos returns editor window coordinates in play mode
## Description
When playing in the editor, `Inputs.GetMousePos()` returns the cursor position relative to the whole editor window instead of the Game View.

It uses `InputManager::GetMousePosition`, which reads the cursor position of the editor window, while the game is rendered inside the Game View panel. The Scene View already removes the panel offset from the mouse position for picking and gizmos, but nothing does it for scripts. In a built game, the window and the rendered area are the same, so the position is correct there.

Any script using the mouse position against what is rendered (UI, crosshair, custom picking) is offset in the editor.

## To Reproduce
1. Create a script with the following content:
```lua
local MousePrinter = {}

function MousePrinter:OnUpdate(deltaTime)
    if Inputs.GetMouseButtonDown(MouseButton.BUTTON_LEFT) then
        local position = Inputs.GetMousePos()
        Debug.Log("Mouse position: " .. position.x .. ", " .. position.y)
    end
end

return MousePrinter
```
2. Add the script to any actor of the scene
3. Press Play
4. Click near the top-left corner of the Game View
5. See that the logged position is far from 0, 0 (it matches the position of the Game View in the editor window)

## Expected behavior
While playing in the editor, `Inputs.GetMousePos()` returns the position relative to the top-left corner of the Game View, the same way it works in a built game.

## Screenshots
N/A

### LINKED ISSUE #846: Expose the size of the rendered area to Lua
Scripts have no way to know the size of the area the game is rendered into. In a built game this is the window, and in the editor it's the Game View, which usually doesn't match the resolution set in the project settings.

Anything working in screen space needs this size : converting the mouse position, placing things relative to the screen edges, or matching the canvas size computed by the engine. The only option right now is to hardcode a resolution, which is wrong in the editor and also breaks as soon as the game window is resized (#837).

## Expected solution
A Lua function returning the size of the rendered area, for example `Inputs.GetViewportSize()` next to `Inputs.GetMousePos()`:
- in a built game, it returns the window size, which is already the render size (`Game.cpp`)
- in the editor, it returns the size used to render the Game View (`AView::GetSafeSize`)

This goes with #845, since the Game View already knows its position and size : the mouse offset and this size could come from the same place.

Gopmyc:
## Description
Makes `Inputs.GetMousePos()` relative to the Game View while playing in the editor, and adds `Inputs.GetViewportSize()`, returning the size of the area the game is rendered into:
- Editor: the Game View mouse position (`AView::GetMousePosition`) and size (`AView::GetSafeSize`), provided by `GameView` through `InputHelpers`.
- Built game: the window mouse position and size, used when no provider is set.

`SceneView` now uses `AView::GetMousePosition` for gizmos and picking instead of duplicating the conversion.

## Related Issue(s)
Fixes #845
Fixes #846

## Review Guidance
- Uses `Window::GetSize` rather than `GetFramebufferSize`, to stay in cursor coordinates on HiDPI screens.
- A closed, collapsed or hidden Game View returns its last known size, like `DebugSceneRenderer`. The size is 0 if the view was never shown or the game window is minimized.

## Screenshots/GIFs
N/A

## AI Usage Disclosure
N/A

## Checklist


adriengivry (approved):
LGTM, code and functionality is sound!

Before merging we should update the description. My understanding is that this PR fixes 2 issues:
#845 and #846.

Some sentence don't make sense to me:
> It reuses the InputHelpers provider pattern from https://github.com/Overload-Technologies/Overload/issues/845

#845 being an issue, there is no reusing of a pattern.

PR #854, opened 2026-09-19
Fix lost key/mouse events and scroll accumulation within a frame (#847)

### LINKED ISSUE #847: Clicks, key presses and wheel notches get lost within a frame
## Description
`InputManager` only keeps the last state of each key and mouse button for the current frame. When a button is pressed and released before the next update, only the release is left, so `Inputs.GetMouseButtonDown` and `Inputs.GetKeyDown` never return true for that press.

The mouse wheel has a similar issue: `OnMouseScroll` replaces the offset instead of adding to it, so several notches received during the same frame count as one.

It's easy to miss at a high frame rate, but a quick tap on a touchpad or a fast wheel spin is enough to trigger it, and it gets worse when the frame rate drops.

## To Reproduce
1. Create a script with the following content:
```lua
local InputLogger = {}

function InputLogger:OnUpdate(deltaTime)
    if Inputs.GetMouseButtonDown(MouseButton.BUTTON_LEFT) then
        Debug.Log("Left button down")
    end

    if Inputs.GetMouseButtonUp(MouseButton.BUTTON_LEFT) then
        Debug.Log("Left button up")
    end

    local scroll = Inputs.GetMouseScroll()

    if scroll.y ~= 0 then
        Debug.Log("Scroll: " .. scroll.y)
    end
end

return InputLogger
```
2. Add the script to any actor of the scene
3. Press Play
4. Tap quickly on a touchpad a few times, then spin the mouse wheel fast
5. See that some taps only log "Left button up", and that fewer scroll values are logged than wheel notches

## Expected behavior
Every press is reported by `GetMouseButtonDown` and `GetKeyDown`, even when the release happens in the same frame, and the scroll offsets received during a frame are added together.

## Screenshots
N/A

Element119rd:
## Description
Fixes the input handling bug reported in #847, where clicks, key presses, and mouse wheel notches get lost within a single frame. 

### Changes
-- Replaced single state event maps in InputManager with separate per-frame flag maps (m_keyPressed, m_keyReleased, m_mouseButtonPressed, m_mouseButtonReleased).
-- Changed OnMouseScroll to accumulate deltas (+=) instead of overwriting.
-- Updated ClearEvents() to reset scroll vector and clear frame flags.
-- Kept public API signature unchanged.

## Related Issue(s)
Fixes #847

## Review Guidance
Main changes are in InputManager.cpp / .h.
ClearEvents() logic remains tied to the end-of-frame tick,

### How to test
1. Build the project.
2. Add the repro script from the issue to any actor, press Play.
3. Tap quickly on a touchpad a few times , each tap should log both `Left button down` and `Left button up`.
4. Spin the mouse wheel fast , logged scroll values should reflect the sum of notches received during the frame, 
5. verify normal held keys/buttons still behave correctly.

## Screenshots/GIFs
none!

## AI Usage Disclosure
none!

## Checklist


Sources/OvWindowing/include/OvWindowing/Inputs/InputManager.h
```diff
@@ -101,8 +101,12 @@ namespace OvWindowing::Inputs
 		OvTools::Eventing::ListenerID m_mouseButtonReleasedListener;
 		OvTools::Eventing::ListenerID m_mouseScrollListener;
 
-		std::unordered_map<EKey, EKeyState> m_keyEvents;
-		std::unordered_map<EMouseButton, EMouseButtonState> m_mouseButtonEvents;
+		std::unordered_map<EKey, bool> m_keyPressed;
```
adriengivry: Since we never use explicit key storing with false, we could use an `std::unordered_set` instead, which will also allow for faster access with `std::unordered_set::contains()`.

This applies to other events as well.

adriengivry (changes requested):

Sources/OvWindowing/include/OvWindowing/Inputs/InputManager.h
Paaanaaamaaa: done. I applied it to all four perframe containers.

Sources/OvWindowing/include/OvWindowing/Inputs/InputManager.h
```diff
@@ -1,12 +1,13 @@
 /**
+/This is InputManager.h
```
adriengivry: ? 😅

Sources/OvWindowing/src/OvWindowing/Inputs/InputManager.cpp
```diff
@@ -1,4 +1,5 @@
 /**
+/This is InputManager.cpp
```
adriengivry: ? 😅

Paaanaaamaaa: I was just changing some file names so i left this comment on them.

Sources/OvWindowing/include/OvWindowing/Inputs/InputManager.h
Paaanaaamaaa: that was a leftover i forgot to remove.😅

adriengivry (approved):

